PR #2, opened 2023-07-24
first public dbt commit

PerriLucas:
First commit to the repository, aimed at pushing the source tables, stg tables and the two mart tables that will be hosted on the public dataset. 

TO DO:

- Name the deduplicated history_assets model so that it does not conflict with the raw table (should it keep being history_assets while the raw table is called history_assets_staging?)
- Add README
- New table additions (?)

### DIFF
--- a/example_profiles.yml
+++ b/example_profiles.yml
@@ -1,0 +1,18 @@
+elementary:
+  target: 'prod'
+  outputs:
+    default:
+      type: bigquery
+
+      # Service acount auth #
+      method: service-account
+      keyfile: {{/path/to/bigquery/keyfile.json}}
+      
+      project: 'test-hubble-319619'
+      dataset: elementary
+      maximum_bytes_billed: 5500000000
+      job_execution_timeout_seconds: 300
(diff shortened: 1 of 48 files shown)


models/marts/enriched_history_operations.sql
```diff
@@ -0,0 +1,355 @@
+{{ config(
+    tags = ["partnership_assets"],
```
chowbao, line 2: nit: remove this tag

example_profiles.yml
```diff
@@ -0,0 +1,18 @@
...
+  target: 'prod'
+  outputs:
+    default:
+      type: bigquery
+
+      # Service acount auth #
+      method: service-account
+      keyfile: {{/path/to/bigquery/keyfile.json}}
+      
+      project: 'test-hubble-319619'
```
chowbao, line 11: nit: remove actual test hubble name

chowbao (approved):

sydneynotthecity (approved):
Agree, looks good and ready to be merged

PR #6, opened 2023-10-26
Some fixes and addition of fee_stats and trade_agg

PerriLucas:
- Fixed materialization for staging, added config for intermediate models. 
- Fixed schema name for staging tables and sources
- Added macro for percentile_iteration
- Added models for fee_stats daily aggregation
- Added models for trade_agg calculation
- Added documentation for models

### DIFF
--- a/models/docs/marts/trade_agg.md
+++ b/models/docs/marts/trade_agg.md
@@ -1,0 +1,301 @@
+[comment]: < Trade Aggregations -
+
+{% docs trade_agg %}
+
+This table contains the aggregated values for the trades involving different pairs of selling/buying assets, over different time periods in the past.
+
+{% enddocs %}
+
+{% docs trade_day_agg %}
+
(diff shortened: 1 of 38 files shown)


models/docs/marts/trade_agg.md
```diff
@@ -0,0 +1,301 @@
+[comment]: < Trade Aggregations -
+
+{% docs trade_agg %}
+
+This table contains the aggregated values for the trades involving different pairs of selling/buying assets, over different time periods in the past.
+
+{% enddocs %}
+
+{% docs trade_day_agg %}
```
sydneynotthecity, line 9: nit: this column is just `day_agg`

sydneynotthecity (approved):
LGTM

models/docs/marts/trade_agg.md
PerriLucas: The doc name is only relevant for when calling it from the .yml file, so there's no issue with it being different from the actual column name. This was called a specific name to keep the custom description.

PR #7, opened 2023-11-21
implement tagging in ci #major

cayod:
With the goal of using dbt-public as a package for stellar-dbt we implement this tag feature in ci. This feature automatically bump and tag master, on merge, with the latest SemVer formatted version.

Every PR that is merged will trigger the CI that will bump tag with patch version unless the merge commit message contains #major or #minor. If two or more are present, the highest-ranking one will take precedence. If #none is contained in the merge commit message, it will skip bumping regardless DEFAULT_BUMP.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+name: Bump version
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - master
+      - main
+
+jobs:
+  build:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+    steps:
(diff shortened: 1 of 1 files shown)


sydneynotthecity (approved):
LGTM

sydneynotthecity:
Is it possible for us to add a check that validates whether a commit message has one of the three tags? (major, minor or patch)? It would be great if we could validate the commit message prior to pushing a commit.

cayod:
> Is it possible for us to add a check that validates whether a commit message has one of the three tags? (major, minor or patch)? It would be great if we could validate the latest commit message prior to opening a PR.

This IC already checks if one of the commit messages has one of the tags and uses the highest-ranked one. If none are defined, CI will default to a patch

PR #11, opened 2023-12-12
dbt Elementary and tests configuration

enercolini:
Initial commit for reviewing the public sources and marts test configurations.

Confirm that the period defined for testing is as expected. Also confirm that the date column chosen (closed_at, batch_insert_ts, etc) to be used as a reference for the defined periods is correct.

During the execution of the `dbt test` on the configured sources, there were a few test failures. If you want more details and the list of failures @sydneynotthecity just let me know and I can provide it to you.

Also there is a few docs that needs updates. If you want to provide the description I can update it.

### DIFF
--- a/models/marts/enriched_history_operations.yml
+++ b/models/marts/enriched_history_operations.yml
@@ -4,12 +4,10 @@ models:
   - name: enriched_history_operations
     description: Aggregate table for the history operations
     tests:
-      - elementary.freshness_anomalies:
-          timestamp_column: closed_at
-          backfill_days: 7
-          time_bucket:
-            period: hour
-            count: 12
+      - dbt_utils.recency:
+          datepart: hour
+          field: cast(closed_at as datetime)
+          interval: 12
           meta:
             description:
               "Monitors the freshness of your table over time, as the expected time between data updates."
--- a/models/marts/enriched_history_operations.yml
+++ b/models/marts/enriched_history_operations.yml
@@ -10,6 +10,20 @@ models:
           time_bucket:
             period: hour
             count: 12
+          meta:
+            description:
+              "Monitors the freshness of your table over time, as the expected time between data updates."
+      - elementary.dimension_anomalies:
+          timestamp_column: closed_at
+          backfill_days: 90
+          dimensions:
+            - "type"
+          time_bucket:
+            period: day
+            count: 1
+          meta:
+            description:
+              "Monitors the frequency of values in the type column over time."
     columns:
       - name: op_id
         description: '{{ doc("operation_id") }}'
@@ -20,21 +34,14 @@ models:
                         and closed_at < timestamp_trunc(current_timestamp(), day)
           - not_null:
               config:
-                where: batch_run_date > current_datetime - interval 2 day
-          - relationships:
-              to: ref('stg_history_operations')
-              field: op_id
-          #- all_source_table_values_in_destination_table:
-          #    to: ref('stg_history_operations')
-          #    field: op_id
+                where: closed_at > current_timestamp - interval 2 day
 
       - name: op_source_account
         description: '{{ doc("source_account") }}'
         tests:
           - not_null:
               config:
-                where: batch_run_date > current_datetime - interval 2 day
-
+                where: closed_at > current_timestamp - interval 2 day
 
       - name: op_source_account_muxed
         description: '{{ doc("source_account_muxed") }}'
@@ -44,24 +51,28 @@ models:
         tests:
           - not_null:
               config:
-                where: batch_run_date > current_datetime - interval 2 day
-          - relationships:
-              to: ref('stg_history_transactions')
-              field: transaction_id
+                where: closed_at > current_timestamp - interval 2 day
 
       - name: type
         description: '{{ doc("type") }}'
         tests:
           - not_null:
               config:
-                where: batch_run_date > current_datetime - interval 2 day
+                where: closed_at > current_timestamp - interval 2 day
+          - dbt_utils.expression_is_true:
+              expression: '<= 26'
+              config:
+                where: closed_at > current_timestamp - interval 90 day
+              meta:
+                description:
+                  "Test to check if the type of operation added is not greater than 26."
 
       - name: type_string
         description: '{{ doc("type_string") }}'
         tests:
           - not_null:
               config:
-                where: batch_run_date > current_datetime - interval 2 day
+                where: closed_at > current_timestamp - interval 2 day
 
       - name: account
         description: '{{ doc("details_account") }}'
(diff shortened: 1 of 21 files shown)


sydneynotthecity:
> there were a few test failures. If you want more details and the list of failures @sydneynotthecity just let me know and I can provide it to you.

Yes, if you can send me those failures that would be great. Feel free to send through Slack or email

models/marts/enriched_history_operations.yml
```diff
@@ -3,12 +3,24 @@ version: 2
...
+          - unique:
+              config:
+                where: closed_at >= timestamp_trunc(timestamp_sub(current_timestamp(), interval 2 day), day) 
+                        and closed_at < timestamp_trunc(current_timestamp(), day)
+          - not_null:
+              config:
+                where: batch_run_date > current_datetime - interval 2 day
           - relationships:
               to: ref('stg_history_operations')
               field: op_id
```
sydneynotthecity, line 26: this is potentially a very expensive query to run if there are no date filters on the `stg_history_operations` table. Do you know if we have the ability to add a where clause that would limit the join?

models/marts/enriched_history_operations.yml
```diff
@@ -19,28 +31,37 @@ models:
...
+                where: batch_run_date > current_datetime - interval 2 day
           - relationships:
               to: ref('stg_history_transactions')
               field: transaction_id
 
       - name: type
         description: '{{ doc("type") }}'
         tests:
-          - not_null
+          - not_null:
```
sydneynotthecity, line 55: Could we add a test that confirms that the `type <= 26`?

models/marts/history_assets.yml
```diff
@@ -3,34 +3,70 @@ version: 2
...
   - name: history_assets
     description: '{{ doc("history_assets_staging") }}'
+    tests:
+      - elementary.freshness_anomalies:
+          timestamp_column: batch_run_date
+          backfill_days: 7
+          time_bucket:
+            period: day 
+            count: 2
+      - dbt_utils.unique_combination_of_columns:
```
sydneynotthecity, line 13: I believe there is an `asset_id` on this table that _should_ be a unique id.

models/marts/history_assets.yml
```diff
@@ -3,34 +3,70 @@ version: 2
...
+          combination_of_columns:
+            - batch_run_date
+            - asset_type
+            - asset_code
+            - asset_issuer
+          config:
+            where: day_agg >= datetime_trunc(datetime_sub(current_datetime(), interval 2 day), day)
```
sydneynotthecity, line 20: update to `batch_run_date`, there is no `day_agg` on `history_assets_staging`

sydneynotthecity (commented):
Couple nits that need to be addressed before approving. 

For the `src_*` tables, it might also be worthwhile to investigate how the query performs with `batch_insert_ts` as the column for the freshness timestamp column. I have no idea if it will do a full table scan since that column is not the partition column, but I would like confirmation that it will not scan the entire table. 

There are a couple of other tables where the date column configured is not the partition key. Given how large these tables are, can you confirm that BQ is not processing all bytes in the table when running the queries? if it is, the columns will need to be updated to the partition column.

Similar to [this PR](https://github.com/stellar/stellar-dbt/pull/98), i think it would be nice to have some volume alerts too if the volume of the data drastically increased or decreased during a timespan. If you are able to implement volume alerts in the other PR, please also apply here.

models/marts/enriched_history_operations.yml
enercolini: Although this test has been removed, I think we could use a where clause to limit the join. We just need to add "config: where:  " and the statement. More information on https://docs.getdbt.com/reference/resource-configs/where

models/marts/enriched_history_operations.yml
enercolini: Implemented through the dbt utils test "expression_is_true".

enercolini:
Once this tests are approved, I'll move forward to configure the slack alerts as requested.

enercolini:
Hi Syd!

We've made some new changes to the test configuration. We've decided to switch the Elementary's freshness test to native [dbt source freshness](https://docs.getdbt.com/reference/resource-properties/freshness) and [dbt utils recency test](https://github.com/dbt-labs/dbt-utils#recency-source) for models. The Elementary test results are based on the data average, which couldn't meet our requirements for defining how long would be considered not fresh. Despite numerous configuration attempts and research, we can 't set a config to alert us if the data is outdated by more than 12 hours or 2 days. Also during the tests it didn't fail for the test data (dev hubble) —where we use samples within a defined period—where the last date was from last year, yet it didn't register any failure. It only failed for data points that were more than 3 standard deviations away from the average. For example, it failed for claimable balances where the batch_run_date pointed as anomalous  an updated at 2024-01-04 15:00:00+00, 1 hours from the last update. And usually the table is updated within 0.52 hours. That's why we've decided to change this, because we consider the goal to alert if the data is delayed from the current date. If you still want to use the elementary freshness test as well, just let me know and I'll add them again.

Moreover, these native dbt and utils tests are faster and require less processing compared to the Elementary test. If you want to check a failure from the dbt source freshness test, you can check it easily on the source.json file the variables and results. For the dbt utils recency we can copy the compiled code from the target file and paste it on BQ to get the exact data that cause the failure. The rest of the tests remain unchanged as they serve their purpose effectively. I'll also be coordinating with our team to ensure everything runs smoothly. If you have any questions, feel free to ask.

tests/ledger_sequence_increment.sql
```diff
@@ -0,0 +1,33 @@
+{{ config(severity="warn"
```
chowbao, line 1: Where is `tests/ledger_sequence_increment.sql` used?

enercolini: Syd asked if we could add a test to ensure that the sequence is always incrementing, so we don't have a case where max(sequence) of batch_id > max(sequence) of batch_id + 1 (next batch id). Since the packages we're using didn't have a test to check this case, I configured as a dbt singular test. Let me know if this makes sense to you. Thanks!

chowbao:
Wanted to confirm this before approving. 

>For the src_* tables, it might also be worthwhile to investigate how the query performs with batch_insert_ts as the column for the freshness timestamp column. I have no idea if it will do a full table scan since that column is not the partition column, but I would like confirmation that it will not scan the entire table.

Was this ever tested or confirmed? If it does cause problems the current batch_run_date/closed_at will work

enercolini:
> Wanted to confirm this before approving.
> 
> > For the src_* tables, it might also be worthwhile to investigate how the query performs with batch_insert_ts as the column for the freshness timestamp column. I have no idea if it will do a full table scan since that column is not the partition column, but I would like confirmation that it will not scan the entire table.
> 
> Was this ever tested or confirmed? If it does cause problems the current batch_run_date/closed_at will work

As I remember, I don't think using batch_insert_ts processed much more than batch_run_date due to the query elementary builds to analyze the metrics. I compared configuring two tests for the same table (account_signers) using different columns and both processed the same size. But for performance matters, I've changed the timestamp column to the partition column for all tables (basically batch_run_date and closed_at). Let me know if this makes sense to you as well. Thank you,

chowbao (approved):

PR #15, opened 2024-01-23
Pass asset info in trade agg

chowbao:
* Pass asset_code, asset_issuer, and asset_type through trade agg instead of join with history_assets
* trade_agg starts with yearly and left joins with other int_* tables in order to preserve all possible assets for trades
Note: joins do not need the above columns. Joining on asset_id does effectively the same thing


sydneynotthecity (approved):
LGTM

PR #17, opened 2024-01-29
Update soroban pipelines

chowbao:
Add Soroban pipelines and changes
* Added source/stg tables for new soroban tables
* Updated `stg_history_operations`, `stg_history_effects`, and `stg_history_transactions` with new soroban columns
* Updated enriched history operations tables to include new soroban columns
* Update docs
* Add ttl_current table

### DIFF
--- a/models/docs/sources/config_settings.md
+++ b/models/docs/sources/config_settings.md
@@ -2,6 +2,9 @@
 
 {% docs config_settings %}
 The config settings table contains the configuration information for Soroban contract resources.
+These settings are set at the network level and requires a validator vote to update.
+[XDR defining the config settings](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-contract-config-setting.x)
+[Fees and metering docs](https://soroban.stellar.org/docs/soroban-internals/fees-and-metering#cost-parameters)
 {% enddocs %}
 
 {% docs config_setting_id %}
--- a/models/docs/sources/config_settings.md
+++ b/models/docs/sources/config_settings.md
@@ -1,0 +1,194 @@
+[comment]: < Config Settings -
+
+{% docs config_settings %}
+The config settings table contains the configuration information for Soroban contract resources.
+{% enddocs %}
+
+{% docs config_setting_id %}
+Config setting id types
+
+| id | Type                                                         |
+|----|--------------------------------------------------------------|
+| 0  |ConfigSettingIdConfigSettingContractMaxSizeBytes              |
+| 1  |ConfigSettingIdConfigSettingContractComputeV0                 |
+| 2  |ConfigSettingIdConfigSettingContractLedgerCostV0              |
+| 3  |ConfigSettingIdConfigSettingContractHistoricalDataV0          |
+| 4  |ConfigSettingIdConfigSettingContractEventsV0                  |
+| 5  |ConfigSettingIdConfigSettingContractBandwidthV0               |
+| 6  |ConfigSettingIdConfigSettingContractCostParamsCpuInstructions |
+| 7  |ConfigSettingIdConfigSettingContractCostParamsMemoryBytes     |
+| 8  |ConfigSettingIdConfigSettingContractDataKeySizeBytes          |
+| 9  |ConfigSettingIdConfigSettingContractDataEntrySizeBytes        |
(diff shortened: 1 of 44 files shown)


models/marts/enriched_history_operations.sql
```diff
@@ -188,6 +192,14 @@ with
             , batch_id
             , batch_run_date
             , batch_insert_ts
+            , asset_balance_changes
+            , parameters
+            , `function`
```
sydneynotthecity, line 197: are we missing function type? Which I know is an overloaded field at this point. Talking about the field with domain values `invoke_contract`, `restore_footprint`, `upload_wasm` etc

models/marts/ledger_current_state/ttl_current.sql
```diff
@@ -0,0 +1,53 @@
...
+            , ttl.ledger_entry_change
+            , ttl.closed_at
+            , ttl.deleted
+            , ttl.batch_id
+            , ttl.batch_run_date
+            , ttl.batch_insert_ts
+            , row_number()
+                over (
+                    partition by ttl.key_hash
+                    order by ttl.last_modified_ledger desc
```
sydneynotthecity, line 28: Do we not also need to order by `ledger_entry_change`? If an entry is deleted, won't the last_modified_ledger be identical to the prior record?

models/marts/ledger_current_state/ttl_current.yml
```diff
@@ -0,0 +1,78 @@
+version: 2
+
+models:
+  - name: ttl_current
+    description: '{{ doc("ttl_current") }}'
+    tests:
+      - dbt_utils.recency:
+          datepart: hour
```
sydneynotthecity, line 8: does this truncate the timestamp to the hourly level? Like if data was older than 24+ hours, would we get an alert? or is this truly evaluating only the hour part of the datetime field?

models/sources/src_contract_code.yml
sydneynotthecity, line 1: I thought these new state tables also had `closed_at` on these tables. Is that not the case?

models/sources/src_history_effects.yml
sydneynotthecity, line 1: Were there any effects related to asset balance changes?

models/staging/stg_history_operations.sql
```diff
@@ -117,10 +117,19 @@ with
             , details.reserve_b_min_amount
             , details.reserve_b_withdraw_amount
             , details.shares
+            , details.asset_balance_changes
+            , details.parameters
+            , details.function
+            , details.address
+            , details.type as soroban_operation_type
```
sydneynotthecity, line 124: ah, disregard my comment above

sydneynotthecity (approved):
Looks good, just had some questions for my own understanding of the data model

models/sources/src_history_effects.yml
chowbao: Yes but they already existed in history_effects. The related fields would be `details.asset_code, details.asset_issuer, details.asset_type, details.amount, details.contract, details.address, details.address_muxed`

models/sources/src_contract_code.yml
chowbao: Ah they do. It's just missing in the yml

models/docs/sources/config_settings.md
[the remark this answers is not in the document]
chowbao: I don't think there is any documentation. I can just link to the xdr and fees and metering doc for now

models/marts/ledger_current_state/ttl_current.sql
chowbao: Ah you're right

models/marts/ledger_current_state/ttl_current.yml
chowbao: The truncate should be fine because datetime still has the hh:mm:ss part of the timestamp. I'm not sure about the 24+ hours alert. I would hope that dbt_utils is smart enough to take the datetime_diff instead of the extract(hour) - extract(hour)

We should probably test it but also we would get an error before this would be a problem because the _current table are run every 30 mins so we would hit the 12 hour recency test regardless

sydneynotthecity: sounds good. let's test and adjust appropriately

PR #20, opened 2024-02-15
Fix test errors in source tables

chowbao:
Casting closed_at to datetime and fixing mismatching time data types

### DIFF
--- a/models/sources/src_history_operations.yml
+++ b/models/sources/src_history_operations.yml
@@ -9,7 +9,7 @@ sources:
         tests:
           - dbt_utils.recency:
               datepart: hour
-              field: closed_at
+              field: cast(closed_at as datetime)
               interval: 12
               config:
                 severity: warn
--- a/models/sources/src_history_operations.yml
+++ b/models/sources/src_history_operations.yml
@@ -9,7 +9,7 @@ sources:
         tests:
           - dbt_utils.recency:
               datepart: hour
-              field: cast(closed_at as datetime)
+              field: batch_run_date
               interval: 12
               config:
                 severity: warn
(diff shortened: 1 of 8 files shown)


models/sources/src_history_operations.yml
```diff
@@ -9,7 +9,7 @@ sources:
         tests:
           - dbt_utils.recency:
               datepart: hour
-              field: closed_at
+              field: cast(closed_at as datetime)
```
sydneynotthecity, line 12: does it matter that the partition column for this table is actually `batch_run_date`?

models/sources/src_history_transactions.yml
```diff
@@ -22,25 +22,25 @@ sources:
             tests:
               - unique:
                   config:
-                    where: closed_at >= datetime_trunc(datetime_sub(current_datetime(), interval 2 day), day)
-                            and closed_at < datetime_trunc(current_datetime(), day)
+                    where: cast(closed_at as datetime) >= datetime_trunc(datetime_sub(current_datetime(), interval 2 day), day)
+                            and cast(closed_at as datetime) < datetime_trunc(current_datetime(), day)
```
sydneynotthecity, line 26: might need to add the partition fields to this where clause otherwise it's gonna be expensive

sydneynotthecity (approved):
Had a question but doesn't block the merge. I think we'll need to query processing closely after deploy to make sure we're not scanning too much data

models/sources/src_history_operations.yml
chowbao: Hmm you're right it doesn't matter

models/sources/src_history_transactions.yml
chowbao: Yeah I'll look at these tables again. It should be based on the partitioned field

PR #32, opened 2024-03-26
Add asset_id missing columns to both trustlines and offers stagings

caioribeiro99:


sydneynotthecity (approved):
Looks good once the GH actions pass

PR #38, opened 2024-03-27
Pre commit branch

laysabit:
same as stellar-dbt for pre-commit, and passing the environment changes that happened in stellar-dbt to stellar-dbt-public.


sydneynotthecity (approved):
Looks good. If you apply any changes in https://github.com/stellar/stellar-dbt/pull/151 , just make sure to propagate here so that the pre commit configs match.

PR #39, opened 2024-03-27
Add account_creation_date and min_sequence_number

vitorweiss:
Added account_creation_date and min_sequence_number to accounts_current table to fix the problem with modeling shown in dbt project evaluator.

### DIFF
--- a/models/marts/ledger_current_state/accounts_current.sql
+++ b/models/marts/ledger_current_state/accounts_current.sql
@@ -59,7 +59,7 @@ with
                 > (select max(t.upstream_insert_ts) from {{ this }} as t)
         {% endif %}
     )
-    
+
     , account_date as (
         select
             account_id
--- a/models/marts/ledger_current_state/accounts_current.sql
+++ b/models/marts/ledger_current_state/accounts_current.sql
@@ -59,10 +59,31 @@ with
                 > (select max(t.upstream_insert_ts) from {{ this }} as t)
         {% endif %}
     )
+    
+    , account_date as (
+        select
+            account_id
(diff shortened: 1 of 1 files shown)


models/marts/ledger_current_state/accounts_current.sql
```diff
@@ -59,10 +59,32 @@ with
...
+        group by account_id
+    )
+
+    , get_creation_account as (
+        select
+            current_accts.*
+            , account_date.account_creation_date
+            , account_date.min_sequence_number
+        from current_accts
+        left join account_date
```
sydneynotthecity, line 78: since the `current_accts` table is built from the `stg_accounts`, I don't think there will be a case when an `account_id` exists in `current_accts` and not `account_date`.  Do you think this could be updated to an `inner join`?

sydneynotthecity (approved):

models/marts/ledger_current_state/accounts_current.sql
vitorweiss: Yeah, I believe the result would not be different, just used the left join because of good practice since the inner join is used more to filter tables. Should I change it?

sydneynotthecity: @vitorweiss yeah, if you can change this to an inner join that would be great. if there ever was a case where a record existed in the `current_accts` table that wasn't in the upstream `stg_accounts`, I would rather drop the record as invalid

sydneynotthecity (approved):

sydneynotthecity:
Closing in favor of #61

PR #48, opened 2024-06-11
Singular tests tag to set Airflow DAG

eduardo-nercolini:
<details>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [ ] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with the jira ticket associated with the PR.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.

### Release planning

* [ ] I've decided if this PR requires a new major/minor/patch version accordingly to
</details>

### What

Added "singular_test" tag to the singular data tests in dbt so they can be used in the Airflow DAG configuration, allowing these tests to run at a defined frequency.

### Why

As requested, the singular tests must run at a different frequency from the current model's build DAGs. For this purpose, a tag has been added so that the DAG and task could be designed using this tag

### Known limitations

N/A


tests/anomaly_detection_trade_count.sql
```diff
@@ -1,4 +1,8 @@
-{{ config(severity="warn") }}
+{{ config(
+    severity="warn"
```
chowbao, line 2: Just confirming: will the `warn` severity still alert in the slack channel? Otherwise we might have to increase the severity to get the alert in slack

eduardo-nercolini: Yes, it continues to send alerts even though the severity is configured as warn. This was the case with an alert that was triggering every 30 minutes, and we had to set up a suppression interval to send it only once a day. Here's an example of such an alert for one of these tests: https://indicium-external.slack.com/archives/C06KR1UB6TG/p1716595711802759

chowbao (approved):

PR #51, opened 2024-06-12
dbt source freshness applied to public sources

eduardo-nercolini:
<details>
  
### PR Structure

* [X] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [X] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [ ] This PR's title starts with the jira ticket associated with the PR.

### Thoroughness

* [ ] This PR adds tests for the most critical parts of the new functionality or fixes.

### Release planning

* [X] I've decided if this PR requires a new major/minor/patch version accordingly to
</details>

### What

This PR brings a new proposal for testing sources using the native dbt test `dbt source freshness`, along with intervals defined according to the build execution (dbt_enriched_base_tables DAG) interval of these models/sources. Thus, data with a severity of 'warn' will be considered outdated if it is more than 30 minutes old, and 'error' for intervals exceeding 2x the interval, meaning a difference greater than 60 minutes.

### Why

Source testing is necessary in the pipeline to identify if the data is not outdated and prevent models from being built based on such data, ensuring greater quality and reliability.

### Known limitations

The suggested interval should be discussed to verify if it is more appropriate and fits the expected intervals according to Stellar data pipeline, in order to properly and more accurately assess whether the data is outdated or not.

### DIFF
--- a/models/sources/src_config_settings.yml
+++ b/models/sources/src_config_settings.yml
@@ -6,6 +6,10 @@ sources:
     tables:
       - name: config_settings
         description: '{{ doc("config_settings") }}'
+        freshness:
+          warn_after: {count: 30, period: minute}
+          error_after: {count: 60, period: minute}
+        loaded_at_field: closed_at
         tests:
           - dbt_utils.unique_combination_of_columns:
               combination_of_columns:
--- a/models/sources/src_accounts.yml
+++ b/models/sources/src_accounts.yml
@@ -6,16 +6,11 @@ sources:
     tables:
       - name: accounts
         description: '{{ doc("accounts") }}'
+        freshness:
+          warn_after: {count: 30, period: minute}
+          error_after: {count: 60, period: minute}
+        loaded_at_field: closed_at
         tests:
-          - dbt_utils.recency:
-              datepart: hour
-              field: closed_at
-              interval: 12
-              config:
-                severity: warn
-              meta:
-                description:
-                  "Monitors the freshness of your table over time, as the expected time between data updates."
           - dbt_utils.unique_combination_of_columns:
               combination_of_columns:
                 - account_id
--- a/models/sources/src_accounts_signers.yml
+++ b/models/sources/src_accounts_signers.yml
@@ -6,16 +6,11 @@ sources:
     tables:
       - name: account_signers
         description: '{{ doc("accounts_signers") }}'
+        freshness:
+          warn_after: {count: 30, period: minute}
+          error_after: {count: 60, period: minute}
+        loaded_at_field: closed_at
         tests:
-          - dbt_utils.recency:
-              datepart: hour
-              field: closed_at
-              interval: 12
-              config:
-                severity: warn
-              meta:
-                description:
-                  "Monitors the freshness of your table over time, as the expected time between data updates."
           - dbt_utils.unique_combination_of_columns:
               combination_of_columns:
                 - account_id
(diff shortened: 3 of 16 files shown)


sydneynotthecity:
@harsha-stellar-data can you take a crack at reviewing this PR?

models/sources/src_config_settings.yml
sydneynotthecity, line 1: This table refreshes on an ad hoc basis. We should not have a freshness test for `config_settings`

eduardo-nercolini: Done.

eduardo-nercolini:
@sydneynotthecity when I ran the freshness tests at these defined intervals, the sources `src_contract_code` and `src_contract_data` resulted in `ERROR STALE`, so I didn't include the test for this sources in this commit. I would like to verify with you if the defined intervals make sense and if indeed the source is outdated, or if, in these cases, a different interval should be defined. I await your feedback to make the config and be able to pass in a new commit. Thanks

eduardo-nercolini:
Ps.: I tested again now and `src_contract_data` passed, but `src_contract_code` didn't. Anyway, I would appreciate your comment on this. Thanks

sydneynotthecity:
> when I ran the freshness tests at these defined intervals, the sources src_contract_code and src_contract_data resulted in ERROR STALE

@enercolini yes, it's fine to remove the freshness test for `src_contract_code`. I'm surprised that `contract_data` failed. @harsha-stellar-data can you profile the `contract_data` table by `batch_run_date` and see if there is an appropriate, predictable interval for a freshness test on this table?

harsha-stellar-data:
@eduardo-nercolini I reviewed the changes are here are some comments
My thoughts - since Freshness is more to do about Source, we should be using closed_at as and when available.

- For history_trades, i see batch_run_date being used but I see "ledger_closed_at" available and is getting frequently and properly modified. Should we be using this timestamp field instead? 

- We do not need Freshness or Recency check on contract_code as the intervals are very unpredicatable

- For Contract_Data, Freshness check at 30 (for Warning), 60 (For Error) should work. I performed the following analysis on the table and did not notice any deviation from the pattern
`WITH src_contract_data AS (
  SELECT
    batch_id,
    MIN(closed_at) AS min_closed,
    MAX(closed_at) AS max_closed
  FROM
    `crypto-stellar.crypto_stellar.contract_data`
  GROUP BY
    batch_id
)
SELECT
  batch_id,
  min_closed,
  max_closed
FROM
  src_contract_data
WHERE
  TIMESTAMP_DIFF(max_closed, min_closed, MINUTE) > 30;
`
- I only noticed Recency check on Contract_Data in the code though. Did we back out the changes based on the failures observed and can you also provide details on what field is being used for checking? We should be using Closed_At but Batch_Run_Date should also work. 

Let me know your thoughts here.

@sydneynotthecity @chowbao

sydneynotthecity:
@harsha-stellar-data I agree, we can update all freshness checks to use `closed_at` so that they're consistent and use the right source column. This aligns with our overall strategy to repartition all tables by `closed_at`

Nice analysis, @eduardo-nercolini let's add the freshness test for `contract_data` back in. if you see failures with the test, can you post the results? There were early gaps in the data when Soroban first launched, but there should not be recent examples of freshness problems

harsha-stellar-data (approved):
Updated changes to check Freshness look good.

PR #52, opened 2024-06-19
[HUBBLE - 466] - feature/docs_and_tests_coverage

debora-hcalves:
<details>
  
### PR Structure

* [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
* [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs
  otherwise).
* [x] This PR's title starts with the jira ticket associated with the PR.

### Thoroughness

* [x] This PR adds tests for the most critical parts of the new functionality or fixes.

### Release planning

* [x] I've decided if this PR requires a new major/minor/patch version accordingly to
</details>

### What

Add tests and documentation to the tables that are missing them.

### Why

To reach the minimum required target (< 50%) of tests and documentation in the public repository so that the package can be included in the dbt project evaluator.

### Known limitations

[N/A]


LucasDias-Indicium (approved):
Great job, Débora.

eduardo-nercolini (approved):

models/marts/ledger_current_state/contract_data_current.yml
```diff
@@ -30,51 +30,120 @@ models:
...
+          - not_null:
+              config:
+                where: closed_at > current_timestamp - interval 2 day
 
       - name: balance_holder
         description: '{{ doc("balance_holder") }}'
+        tests:
+          - not_null:
+              config:
+                where: closed_at > current_timestamp - interval 2 day
```
sydneynotthecity, line 71: Does this test actually pass? I think there are blank balance_holders (maybe with empty string?) in the table

models/marts/ledger_current_state/contract_data_current.yml
sydneynotthecity, line 1: i believe any of the balance/balance holder/asset fields are nullable. If that's true, these tests should be removed

models/sources/src_config_settings.yml
sydneynotthecity, line 1: this is a wonky data structure. all fields are nullable except for the `config_setting_id`. probably should remove the non null tests

models/sources/src_contract_data.yml
sydneynotthecity, line 1: i believe these fields can be nullable

models/sources/src_history_trades.yml
sydneynotthecity, line 1: asset_code and asset_issuer are null for `native` asset (XLM)

sydneynotthecity (commented):
Some minor feedback regarding nullability: for the `native` asset type (XLM), both the `asset_code` and the `asset_issuer` are null. I think they're represented with empty spaces, but it's effectively null.

debora-hcalves:
@sydneynotthecity Yes, when I tested in BQ before adding these tests, these columns behaved as empty strings, so in the query, for example, `where asset_code is null`, it didn't return data. Anyway, I added a condition in the test configuration to ignore when the `asset_type` is '**native**' or **3** (depending on the column and the table) and added a few more tests for accepted_values. I also removed the tests for the other columns in `src_config_settings` as you requested.

chowbao (approved):

PR #83, opened 2024-09-03
Move tests from scheduled queries / business queries to DBT

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

This PR:
- changes severity of existing tests to error instead of warning to make it compatible with tests from scheduled queries
- Adds few more tests for column checks based on list of scheduled queries
- Adds tests from scheduled queries as generic tests

### Why

To centralize the data quality tests

Related PR: https://github.com/stellar/stellar-dbt/pull/227

### Known limitations

None

### DIFF
--- a/models/marts/history_assets.yml
+++ b/models/marts/history_assets.yml
@@ -9,9 +9,17 @@ models:
           field: cast(batch_run_date as timestamp)
           interval: 2
           config:
-            severity: warn
+            severity: error
           meta:
             description: "Monitors the freshness of your table over time, as the expected time between data updates."
+      - incremental_unique_combination_of_columns:
+          combination_of_columns:
+            - batch_run_date
+            - asset_type
+            - asset_code
+            - asset_issuer
+          date_column_name: "batch_run_date"
+          greater_than_equal_to: "2 day"
     columns:
       - name: asset_id
         description: '{{ doc("assets_id") }}'
--- a/models/marts/history_assets.yml
+++ b/models/marts/history_assets.yml
@@ -14,7 +14,6 @@ models:
             description: "Monitors the freshness of your table over time, as the expected time between data updates."
       - incremental_unique_combination_of_columns:
           combination_of_columns:
-            - batch_run_date
             - asset_type
             - asset_code
             - asset_issuer
--- a/models/sources/src_accounts.yml
+++ b/models/sources/src_accounts.yml
@@ -14,7 +14,6 @@ sources:
           - incremental_unique_combination_of_columns:
               combination_of_columns:
                 - account_id
-                - sequence_number
                 - ledger_entry_change
                 - last_modified_ledger
               date_column_name: "batch_run_date"
--- a/models/sources/src_accounts.yml
+++ b/models/sources/src_accounts.yml
@@ -14,6 +14,7 @@ sources:
           - incremental_unique_combination_of_columns:
               combination_of_columns:
                 - account_id
+                - sequence_number
                 - ledger_entry_change
                 - last_modified_ledger
               date_column_name: "batch_run_date"
--- a/tests/bucketlist_db_size_check.sql
+++ b/tests/bucketlist_db_size_check.sql
@@ -1,0 +1,27 @@
+{{ config(
+    severity="error"
+    , tags=["singular_test"]
+    )
+}}
+
+with bucketlist_db_size as (
+  select sequence,
+    closed_at,
+    total_byte_size_of_bucket_list / 1000000000 as bl_db_gb
+  from `crypto-stellar.crypto_stellar.history_ledgers`
+  where closed_at >= current_timestamp() - interval 1 hour
+  -- alert when the bucketlist has grown larger than 12 gb
+    and total_byte_size_of_bucket_list / 1000000000 >= 12 
+)
+
+select 
+if (
+  (select count(*)
+    from bucketlist_db_size
+  ) = 0,
+  --Return when True:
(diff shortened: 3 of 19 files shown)


models/marts/history_assets.yml
```diff
@@ -9,9 +9,17 @@ models:
           field: cast(batch_run_date as timestamp)
           interval: 2
           config:
-            severity: warn
+            severity: error
           meta:
             description: "Monitors the freshness of your table over time, as the expected time between data updates."
+      - incremental_unique_combination_of_columns:
+          combination_of_columns:
+            - batch_run_date
```
chowbao, line 17: I'm not sure `batch_run_date` should be included. My assumption was that the `history_assets` table was unique on `asset_code, asset_issuer, and asset_type` otherwise we would have "duplicate assets" based where each asset would have multiple `batch_run_dates`

models/sources/src_accounts.yml
```diff
@@ -14,7 +14,6 @@ sources:
           - incremental_unique_combination_of_columns:
               combination_of_columns:
                 - account_id
-                - sequence_number
```
chowbao, line 17: I think `sequence_number` should be included. The src/staging `accounts` table should record every instance of an account on every `sequence_number` iirc

amishas157: Got it. I will keep it then, It was not present in the scheduled query, so I treated that as source of truth

tests/bucketlist_db_size_check.sql
```diff
@@ -0,0 +1,17 @@
+{{ config(
+    severity="error"
+    , tags=["singular_test"]
```
chowbao, line 3: FYI this tag runs every 30 mins in airflow. This is a higher frequency compared to what the cloud function and scheduled query tests used to run at. Which is good.

Just mentioning this in case we get noisy alerts where we might want to adjust the query and/or the frequency the tests are run (possibly with a separate dbt tag).

tests/bucketlist_db_size_check.sql
```diff
@@ -0,0 +1,17 @@
...
+    )
+}}
+
+with bucketlist_db_size as (
+  select sequence,
+    closed_at,
+    total_byte_size_of_bucket_list / 1000000000 as bl_db_gb
+  from {{ source('crypto_stellar', 'history_ledgers') }}
```
chowbao, line 11: I think this should be a `ref('stg_history_ledgers')` instead of a source. Otherwise these tests would be hardcoded to just prod right?

tests/eho_by_ops.sql
```diff
@@ -0,0 +1,36 @@
...
+-- Enriched_history_operations table is dependent on the
+-- history_operations table to load. It is assumed that
+-- any id present in the upstream table should be loaded in
+-- the downstream. If records are not present, alert the team.
+WITH find_missing AS (
+  SELECT op.id,
+    op.batch_run_date,
+    op.batch_id
+  FROM {{ source('crypto_stellar', 'history_operations') }} op
+  LEFT OUTER JOIN {{ ref('enriched_history_operations') }} eho
```
chowbao, line 16: Same comment here

I think this should be a ref('stg_history_operations') instead of a source. Otherwise these tests would be hardcoded to just prod right?

tests/bucketlist_db_size_check.sql
amishas157: Wont test still use `crypto_stellar.history_ledgers` instead of `test_crypto_stellar.history_ledgers` ?

Example: https://github.com/stellar/stellar-dbt-public/blob/2e63a8a7a00937213372d0ab636016373d59f94b/models/staging/stg_account_signers.sql#L6-L28

models/marts/history_assets.yml
amishas157: Addressed in https://github.com/stellar/stellar-dbt-public/pull/83/commits/6bd6abf411f043b3e30af1bfa1780741b9d9cac1

models/sources/src_accounts.yml
amishas157: Addressed in https://github.com/stellar/stellar-dbt-public/pull/83/commits/6bd6abf411f043b3e30af1bfa1780741b9d9cac1

tests/bucketlist_db_size_check.sql
chowbao: It will use the test project/dataset because the source for staging tables will be overwritten by the dbt_project.yml in the private dbt repo
https://github.com/stellar/stellar-dbt/blob/master/dbt_project.yml#L63-L68
```
+project: "{% if target.name == 'prod' %} crypto-stellar {% else %} {{ target.project }} {% endif %}"
```

amishas157: Got it. Yes, agree in that case we should just use ref

amishas157: Addressed in https://github.com/stellar/stellar-dbt-public/pull/83/commits

tests/ledger_sequence_increment.sql
```diff
@@ -11,7 +11,7 @@ with
             , batch_id
             , closed_at
             , max(sequence) as max_sequence
-        from {{ source('crypto_stellar', 'history_ledgers') }}
+        from {{ ref('stg_history_ledgers') }}
```
chowbao, line 14: 🙏

chowbao (approved):

PR #87, opened 2024-09-06
Truncate timestamp to shift windowing in trade agg

sydneynotthecity:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

Truncate the timestamp on date filter to ensure that the full day of data is considered. The current code will only pull partial date for the current date, dependent on runtime

### Why

Jenkins build pipeline is failing tests for `trade_agg` and `int_trade_agg_day` because no data is pulled for the 1 day aggregate. This is because the current filter expects data _newer_ than current timestamp: 

```
where ledger_closed_at < timestamp_add('{{ dbt_airflow_macros.ts() }}',interval 1 day )
and ledger_closed_at >= '{{ dbt_airflow_macros.ts() }}'
```

Added logic to truncate the timestamp to the beginning of the date, so that all data for the current run is pulled.

### Known limitations

This is less of a problem in production because we pass Airflow variables. The jenkins build job relies on the default, current timestamp, instead of an env variable being passed in. 

To make this fix permanent, we also need to pass current_date - 1 in via env var in the `jenkins_lint_and_test` build script.

### DIFF
--- a/models/intermediate/trades/int_trade_agg_month.sql
+++ b/models/intermediate/trades/int_trade_agg_month.sql
@@ -25,8 +25,8 @@ with
             , buying_amount
         from {{ ref('stg_history_trades') }}
         where
-            ledger_closed_at < timestamp_add('{{ dbt_airflow_macros.ts(timezone=none) }}', interval 1 day)
-            and ledger_closed_at >= timestamp_sub('{{ dbt_airflow_macros.ts(timezone=none) }}', interval 31 day)
+            ledger_closed_at < timestamp_add(timestamp_trunc('{{ dbt_airflow_macros.ts() }}', day), interval 1 day)
+            and ledger_closed_at >= timestamp_sub(timestamp_trunc('{{ dbt_airflow_macros.ts() }}', day), interval 31 day)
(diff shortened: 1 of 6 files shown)


models/intermediate/trades/int_trade_agg_month.sql
```diff
@@ -120,7 +120,7 @@ with
     /* obtain aggregate function metrics for the asset pair */
     , trade_day_agg_group as (
         select
-            date('{{ dbt_airflow_macros.ts(timezone=none) }}') as day_agg
+            date('{{ dbt_airflow_macros.ts() }}') as day_agg
```
chowbao, line 123: Just wondering. Is there an issue with using timezones=none? I'd assume that's the same as the default UTC

chowbao (approved):

models/intermediate/trades/int_trade_agg_month.sql
sydneynotthecity: there isn't, but it is the default. My thought was why specify the default

PR #94, opened 2024-09-18
Increase test coverage for marts

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

This PR adds tests for marts. Also, exports model config as meta tag so that we can query it from elementary schema. This is further used to calculate primary test coverage

### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -63,6 +63,11 @@ models:
       tests:
         fct_missing_primary_key_tests:
           +enabled: false
+        fct_sources_without_freshness:
+          +enabled: false
+      documentation:
+        fct_undocumented_sources:
+          +enabled: false
 
 seeds:
   stellar_dbt_public:
--- a/models/marts/enriched_history/enriched_history_operations.sql
+++ b/models/marts/enriched_history/enriched_history_operations.sql
@@ -1,12 +1,17 @@
-{{ config(
-    tags = ["enriched_history_operations"],
-    materialized='incremental',
-    unique_key=["op_id"],
-    partition_by={
+{% set meta_config = {
+    "materialized": "incremental",
+    "unique_key": ["op_id"],
+    "cluster_by": ["ledger_sequence","transaction_id","op_account_id","type"],
+    "partition_by": {
         "field": "closed_at"
         , "data_type": "timestamp"
-        , "granularity": "month"},
-    cluster_by=["ledger_sequence","transaction_id","op_account_id","type"]
+        , "granularity": "month"}
+} %}
+
+{{ config(
+    tags=["enriched_history_operations"],
+    meta=meta_config,
+    **meta_config,
     )
 }}
 
--- a/models/marts/enriched_history/enriched_history_operations.sql
+++ b/models/marts/enriched_history/enriched_history_operations.sql
@@ -5,11 +5,11 @@
     "partition_by": {
         "field": "closed_at"
         , "data_type": "timestamp"
-        , "granularity": "month"}
+        , "granularity": "month"},
+    "tags": ["enriched_history_operations"]
 } %}
 
 {{ config(
-    tags=["enriched_history_operations"],
     meta=meta_config,
     **meta_config,
     )
--- a/models/marts/ledger_current_state/config_settings_current.yml
+++ b/models/marts/ledger_current_state/config_settings_current.yml
@@ -5,9 +5,9 @@ models:
     description: '{{ doc("config_settings_current") }}'
     tests:
       - dbt_utils.recency:
-          datepart: day
-          field: cast(closed_at as timestamp)
-          interval: 2
+          datepart: hour
+          field: closed_at
+          interval: 12
           config:
             severity: '{{ "error" if target.name == "prod" else "warn" }}'
           meta:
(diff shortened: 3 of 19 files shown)


dbt_project.yml
```diff
@@ -64,6 +64,11 @@ models:
       tests:
         fct_missing_primary_key_tests:
           +enabled: false
+        fct_sources_without_freshness:
+          +enabled: false
+      documentation:
+        fct_undocumented_sources:
+          +enabled: false
```
sydneynotthecity, line 71: just curious: are these tests that you would like us to use in the future? (eventually enabled = true) 

or were you getting an error in CI?

models/marts/ledger_current_state/config_settings_current.yml
```diff
@@ -3,7 +3,15 @@ version: 2
...
   - name: config_settings_current
     description: '{{ doc("config_settings_current") }}'
-
+    tests:
+      - dbt_utils.recency:
+          datepart: hour
+          field: closed_at
+          interval: 12
+          config:
+            severity: '{{ "error" if target.name == "prod" else "warn" }}'
```
sydneynotthecity, line 12: config setting updates are rare occurrences. I think we've changed the settings [4 times](https://stellar.expert/explorer/public/protocol-history) since June. We shouldn't have a freshness check on this table as the settings updates are ad hoc

models/marts/ledger_current_state/contract_code_current.yml
```diff
@@ -3,7 +3,15 @@ version: 2
 models:
   - name: contract_code_current
     description: '{{ doc("contract_code_current") }}'
-
+    tests:
+      - dbt_utils.recency:
+          datepart: hour
+          field: closed_at
+          interval: 12
```
sydneynotthecity, line 10: Have you profiled the data to know if this interval window is large enough to not trigger false positives? `contract_code` is another table that is updated infrequently. Might be worthwhile to profile occurrence of writes to this table.

sydneynotthecity: Or update the severity to warn

dbt_project.yml
amishas157: dbt_project_evaluator runs all these checks by default and therefore the CI was failing. We could turn it on later on eventually if we decide to follow best practices by dbt.

Otherwise, we could leave it as is

models/marts/ledger_current_state/config_settings_current.yml
amishas157: Makes sense. In that case, I can make this an exception in our test coverage or can also make threshold really high

models/marts/ledger_current_state/contract_code_current.yml
amishas157: You are right, it does not change frequently. Will increase the thresholds
Screenshot 2024-09-19 at 8 34 08 PM

models/marts/enriched_history/enriched_history_operations.sql
```diff
@@ -1,12 +1,17 @@
-{{ config(
-    tags = ["enriched_history_operations"],
-    materialized='incremental',
-    unique_key=["op_id"],
-    partition_by={
+{% set meta_config = {
```
chowbao, line 1: What's the difference between `config` and `set meta_config`?

chowbao: I see `exports model config as meta tag so that we can query it from elementary schema. This is further used to calculate primary test coverage`

models/marts/enriched_history/enriched_history_operations.sql
```diff
@@ -1,12 +1,17 @@
...
+    "partition_by": {
         "field": "closed_at"
         , "data_type": "timestamp"
-        , "granularity": "month"},
-    cluster_by=["ledger_sequence","transaction_id","op_account_id","type"]
+        , "granularity": "month"}
+} %}
+
+{{ config(
+    tags=["enriched_history_operations"],
```
chowbao, line 12: nit: Why not include tags in the `meta_config` variable if you're unpacking **meta_config anyways?

models/marts/ledger_current_state/config_settings_current.yml
sydneynotthecity: i would just make an exception. We can't reliably predict when updates to config settings occur

dbt_project.yml
sydneynotthecity: ah, ok. In that case leave it as is and we can revisit

models/marts/enriched_history/enriched_history_operations.sql
amishas157: Yeah, I thought about same but did not have a preference. `tags` are anyways part of elementary schema. 
Can export tag as `meta_config` as well if you have a preference

chowbao: I'd prefer it in `meta_config` because that means the `set meta_config` block contains all the config info. And the actual config block only unpacks meta_config

chowbao:
LGTM; Only thing is `contract_code recency test` from [this comment](https://github.com/stellar/stellar-dbt-public/pull/94#discussion_r1766984363)

adding tags to the `meta_config`block is optional

PR #98, opened 2024-09-23
WIP - Feature Branch for Contract Data and Code Staging tables

harsha-stellar-data:
<details>
- Incremental Staging Models for the Contract tables (Data and Code)
-

### PR Structure

      otherwise).
- [X] This PR's title starts with the jira ticket associated with the PR.

### Thoroughness

- [X] This PR adds tests for the most critical parts of the new functionality or fixes.

### Release planning

</details>

### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]


models/staging/stg_soroban_contract_code.sql
```diff
@@ -0,0 +1,191 @@
...
+            , cc.batch_id
+            , cc.batch_run_date
+        from
+            contract_code as cc
+        left join
+            ttl_data as ttl
+            on
+            cc.ledger_key_hash = ttl.key_hash
+        where
+            cc.row_num = 1  -- Only the latest record per ledger_key_hash from contract_code
```
chowbao, line 158: I don't think using `row_number` and taking the latest record is the correct thing to do for an SCD type 2 table. `lead` would probably be more fitting as we want to record every change

models/staging/stg_soroban_contract_code.sql
chowbao, line 1: There shouldn't be any logic in stg_* tables. I recommend moving this logic to an int_* table

models/staging/stg_soroban_contract_code.sql
```diff
@@ -0,0 +1,191 @@
...
+            WHERE DATE(cc.closed_at) = '{{ var("recovery_date") }}'
+
+        -- Date Range recovery
+        {% elif var('recovery_type', 'Range') == 'Range' %}
+            WHERE DATE(cc.closed_at) BETWEEN '{{ var("recovery_start_day") }}' AND '{{ var("recovery_end_day") }}'
+
+        -- Full recovery
+        {% elif var('recovery_type', 'Full') == 'Full' %}
+            -- No WHERE condition for full recovery, as we want all data
+        {% endif %}
```
chowbao, line 76: I'd recommend skipping recovery logic for now. I think it's more important to get a working frontfill/full-refresh solution first

models/dims/contract_details_hist.yml
chowbao, line 1: I'd recommend skipping tests and documentation for now in favor of getting a working solution first.

As an aside, descriptions should point to `models/docs/*.md` files instead of written out in the .yml files

models/staging/stg_soroban_contract_code.sql
```diff
@@ -0,0 +1,191 @@
...
+
+    -- Regular cadence filtering (2-hour forward window, 4-hour backward window)
+    {% else %}
+        WHERE
+            timestamp(cc.closed_at) < timestamp_add(timestamp('{{ var("airflow_start_timestamp") }}'), interval 2 hour)
+            AND timestamp(cc.closed_at) >= timestamp_sub(timestamp('{{ var("airflow_start_timestamp") }}'), interval 4 hour)
+    {% endif %}
```
chowbao, line 83: How did you come up with this default time range? It would provide weird gaps in data ranges depending on the cadence of the task that airflow schedules. Like if the model ran daily, I don't think it'd include all data

models/staging/stg_soroban_contract_code.sql
```diff
@@ -0,0 +1,191 @@
...
+    SELECT COUNT(*) AS tgt_row_count FROM {{ ref('contract_details_hist') }}
+{% endset %}
+
+{% set results = run_query(target_table_query) %}
+
+{% if execute %}
+    {% set is_empty_target_table = (results.columns[0].values()[0] == 0) %}
+{% else %}
+    {% set is_empty_target_table = false %}
+{% endif %}
```
chowbao, line 25: What are you trying to accomplish with this logic?

I think dbt automatically creates the table on the first run and I _think_ that it creates this table with the --full-refresh option

models/staging/stg_soroban_contract_code.sql
sydneynotthecity: Very lightweight logic is permissible in the staging tables as long as it is _generally_ applicable to downstream tables. ie: renaming, type casting, converting cents to dollars, etc (see [staging docs](https://docs.getdbt.com/best-practices/how-we-structure/2-staging#staging-models) for more details)

In this case, a specific date filter and join are applied, which should be moved into [intermediate models](https://docs.getdbt.com/best-practices/how-we-structure/3-intermediate). We have some staging models built for `contract_code` and `contract_data`, that you should be able to leverage for your code

models/staging/stg_soroban_contract_data.sql
chowbao, line 1: Same comments apply to contract_data

models/staging/stg_soroban_contract_code.sql
chowbao, line 1: What is the stg logic is trying to accomplish?

The SQL itself is finding the latest entries of a ledger_key_hash (contract_code) for a given time range window. Can you explain how this will be used to create a dim table?

models/staging/stg_soroban_contract_code.sql
```diff
@@ -0,0 +1,191 @@
...
+            , cc.batch_run_date
+        from
+            contract_code as cc
+        left join
+            ttl_data as ttl
+            on
+            cc.ledger_key_hash = ttl.key_hash
+        where
+            cc.row_num = 1  -- Only the latest record per ledger_key_hash from contract_code
+            and (ttl.row_num = 1 or ttl.row_num is null)  -- Only the latest record from ttl, if available
```
sydneynotthecity, line 159: Is there ever a case when the row_num is null? I would expect it to be not null

sydneynotthecity (commented):
Couple high level design questions I think we need to align on. Both questions are raised in comments in the code:

1. The models are currently `staging` models. To align with dbt best practices, these should be moved to `intermediate` folder.
2. Do we expect the SCD II tables to track _every_ change, or compact changes over a time interval

models/staging/stg_soroban_contract_code.sql
harsha-stellar-data: I opted to go for row_num nd get the latest for each batch run window (2 hrs)  instead of recording each and every change based on this conversation and some profiling
https://stellarfoundation.slack.com/archives/C073MNCNVQQ/p1726585866116489

SCD can record each and every materialistic change that are meaningful to the end users.

Ex: 
`select
*
from
`crypto-stellar.crypto_stellar.contract_code`
where contract_code_hash='4b9316721487281d8201e1c6044544400f120253487971e339eb23a465516935'
order by ledger_key_hash,closed_at
;`

It has ledger entry related values change but the columns that Users would be interested (n_ fields) only changed twice across 10 records. That is also the reason I went with this approach capturing the latest change per each batch window run (which would be 2 hours in this case)

chowbao: Ideally we'd have both a dim table that records every ledger state change as well as a dim table with daily changes. But yeah for now compacting makes sense. 

I would recommend compacting to daily if the most likely use case is daily+ aggregates

models/dims/contract_details_hist.sql
```diff
@@ -0,0 +1,9 @@
+{{ config(materialized = 'table') }}
+
+select
+    'dummy_contract_id' as contract_id
+    , 'dummy_ledger_key_hash' as ledger_key_hash
+    , current_timestamp() as start_date
+    , current_timestamp() as end_date
```
amishas157, line 7: Can we avoid using `current_timestamp()` functions to keep pipelines idempotent? It could be replaced with airflow date `'{{ dbt_airflow_macros.ts(timezone=none) }}'`

Example: 
https://github.com/stellar/stellar-dbt-public/blob/be4de943783abba9c4dbb3a8adaacca8097e30b9/models/marts/ledger_current_state/ttl_current.sql#L33C59-L33C103

models/staging/stg_soroban_contract_data.sql
```diff
@@ -0,0 +1,172 @@
...
+    , sd.closed_at
+    , sd.key_hash
+    , sd.ttl_closed_at
+    , sd.live_until_ledger_seq
+    , sd.ttl_deleted
+    , sd.batch_insert_ts
+    , sd.airflow_start_ts
+    , sd.batch_id
+    , sd.batch_run_date
+    , current_timestamp() as dw_load_ts
```
amishas157, line 168: Maybe use `sd.airflow_start_ts` as dw_load_ts

Related to https://github.com/stellar/stellar-dbt-public/pull/98/files#r1772773579

PR #160, opened 2025-10-23
Add staging asset model

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

This PR adds stg_assets model to have a single source of truth for all assets and associated contract ids. This also joins with other staging models to get asset contract id

Models where contract id was not derived
- stg_contract_data - already has contract id
- stg_history_assets - to be deprecated
- stg_history_effects - to be deprecated
- stg_history_operations - there are many assets column, we need to decide which we want to add asset for

Before merging, I would also need to add contract id to snapshots and downstream models
### Why

### Known limitations

[TODO or N/A]

### DIFF
--- a/models/staging/stg_assets.yml
+++ b/models/staging/stg_assets.yml
@@ -16,6 +16,9 @@ models:
 
       - name: asset_contract_id
         description: '{{ doc("asset_contract_id") }}'
+        tests:
+          - not_null
+          - unique
 
       - name: created_at
         description: '{{ doc("asset_created_at") }}'


models/staging/stg_assets.yml
```diff
@@ -0,0 +1,45 @@
...
+  - name: stg_assets
+    tests:
+      - incremental_unique_combination_of_columns:
+          date_column_name: "closed_at"
+          greater_than_equal_to: "2 day"
+          combination_of_columns:
+            - asset_type
+            - asset_code
+            - asset_issuer
+            - contract_id
```
chowbao: contract_id itself is a unique combination. There is only 1 contract_id per asset

models/staging/stg_trust_lines.yml
```diff
@@ -162,3 +162,8 @@ models:
           - incremental_not_null:
               date_column_name: "closed_at"
               greater_than_equal_to: "2 day"
+
+      - name: asset_contract_id
+        description: '{{ doc("asset_contract_id") }}'
+        tests:
+          - not_null
```
chowbao: Unfortunately I think there are some null contract_ids for trustlines

This happens for assets that are created but have zero events (no mints, burns, transfers, clawbacks).

All these cases have zero balances

This will be resolved whenever we add contract_id to stellar-etl

amishas157:
> LGTM
> 
> Double checking Are you going to make this a table instead of a view in the dbt_project config in the private stellar-dbt?

I think we should and have it build as part of a new task in enriched_base_table DAG

PR #162, opened 2025-10-31
Use batch_start_date and batch_end_date params

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

Add the batch_start_date and batch_end_date parameters for incremental build windowing

### Why

All incremental models should use these new params to make incremental building consistent throughout all our dbt models

### Known limitations

* Currently our airflow dags for current_state run with the assumption that the current state dbt models run with a 7 day lookback window. The airflow dags will need to be updated to run with `catchup=true` or we need to decide if there should be an exception to the current state models to run over a larger window instead of with the airflow batch intervals
* The account_balances models are not modified in this PR. They will be updated in the existing account balances PR


models/intermediate/trades/int_trade_agg_year.sql
```diff
@@ -3,6 +3,10 @@
     )
 }}
 
+-- TODO: The int_trade_agg tables need to be refactored to handle incremental
+-- builds correctly. Currently it only builds a single day even if the model
+-- is full-refreshed.
+
 /* TO-DO This query is costly and can be optimized by
 creating an intermediate source that contains only 1 year of trading history.
 That table could be used as a single source for each aggregate period
```
chowbao: Seems like we already knew the trade_agg model wasn't optimized.

For now we should just update the airflow params and timestamp casting

models/marts/enriched_history/enriched_history_operations.sql
```diff
@@ -41,11 +44,9 @@ with
...
         from {{ ref('stg_history_ledgers') }}
         where
-            cast(batch_run_date as date) < date_add(date('{{ dbt_airflow_macros.ts(timezone=none) }}'), interval 2 day)
-            and date(closed_at) < date_add(date('{{ dbt_airflow_macros.ts(timezone=none) }}'), interval 1 day)
-            {% if is_incremental() %}
-                and cast(batch_run_date as date) >= date_sub(date('{{ dbt_airflow_macros.ts(timezone=none) }}'), interval 1 day)
-                and date(closed_at) >= date_sub(date('{{ dbt_airflow_macros.ts(timezone=none) }}'), interval 1 day)
+            closed_at < timestamp(date('{{ var("batch_end_date") }}'))
+        {% if is_incremental() %}
+                and closed_at >= timestamp(date('{{ var("batch_start_date") }}'))
```
chowbao: I believe the extra date buffering was a safety net. This probably isn't needed anymore

models/marts/enriched_history/enriched_history_operations.sql
sydneynotthecity: can you add an incremental predicate clause to the model config? This model would greatly benefit from that addition

models/marts/enriched_history/enriched_history_operations_soroban.sql
sydneynotthecity: add incremental predicate

chowbao:
> Any `*_current` model should use `closed_at` as the `order by` column in the qualify statement. When we had originally written these models, `closed_at` didn't exist on all the tables. Now that it does, we should be using only that column for ranking.
> 
> Once that change and `incremental predicates` are adding to eho, this is good to release. I don't need to rereview

Hmm but not all the state tables are partitioned on `closed_at`. This would contradict the "use partition column and use the partition column data type to filter" idea

models/marts/ledger_current_state/contract_data_current.sql
```diff
@@ -40,17 +40,27 @@ with
...
-            where
-                TIMESTAMP(cd.closed_at) >= TIMESTAMP_SUB('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL 7 day )
-        {% endif %}
+            -- The extra day date_sub is useful in the case the first scheduled run for a day is skipped
+            -- because the DAG is configured with catchup=false
+            and closed_at >= timestamp(date_sub(date('{{ var("batch_start_date") }}'), interval 1 day))
+    {% endif %}
+        qualify row_number()
+            over (
+                partition by cd.contract_id, cd.ledger_key_hash
```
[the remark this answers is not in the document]
chowbao: Updated in 27e97c047f2c67d7b7f016bcb338ff9ba280bd83

sydneynotthecity:
> Hmm but not all the state tables are partitioned on closed_at. This would contradict the "use partition column and use the partition column data type to filter" idea

I see where we went wrong. My feedback was for the _qualify_ statement, _not_ for the `where` clause. 

You are right, it's an antipattern to filter in a `where` clause on any time field beyond the partition key. This should be reverted to `batch_run_date`. Apologies for the confusion. 

What I want is the qualify statement to use `closed_at` for ordering, instead of a list of columns.

So:
```
qualify row_number()
            over (
                partition by s.account_id, s.signer
                order by s.closed_at desc
            )
        = 1
```

instead of 
```
qualify row_number()
            over (
                partition by s.account_id, s.signer
                order by s.last_modified_ledger desc, s.ledger_entry_change desc
            )
        = 1
```

PR #204, opened 2026-03-04
[#189] Add Soroban core metrics and config settings snapshot models

### LINKED ISSUE #189: Normalize `core_metrics` events into a standalone Soroban utilization table
### What problem does your feature solve?

Currently, calculating Soroban network utilization requires a full table scan of the history_contract_events table (which contains millions of rows) and heavy on-the-fly JSON parsing. Core_metrics topic and values must be decoded in order to make the data useful, but there is no good way to pre-filter the data from the table since parsing of the topic is required to filter on topic. (See this [metabase dashboard](https://hubble-metabase.com/dashboard/125-ledger-level-soroban-limits?threshold=0.7) as an example)

Current Pain Points:
- Latency: Dashboards currently take 2-4 minutes to load, which is a significant barrier to network monitoring.
- Compute Cost: Every query execution incurs a "JSON tax," repeatedly parsing `topics_decoded` and `data_decoded` for the same historical data.
- Scalability: As we push for higher throughput and network performance in 2026, the volume of contract events will grow, making the current ad-hoc parsing method unsustainable.
- Complexity: Users must write complex SQL involving JSON_VALUE and array offsets to access basic metric keys. This will not scale as we migrate this into omni, which requires structured, denormalized data for the best semantic layer performance.

### What would you like to see?

I propose creating a specialized downstream table (e.g., silver.fact_soroban_utilization) that extracts and flattens core_metrics events into a structured format.

**Proposed Table Schema:**
- closed_at (TIMESTAMP): Partitioning column.
- ledger_sequence (INT64): For ledger-grain analysis.
- transaction_hash (STRING): For transaction-grain analysis.
- contract_id (STRING): Clustering column.
- metric_key (STRING): The symbol extracted from topics_decoded[1] (e.g., CPU, RAM, Read/Write bytes).
- metric_value (INT64): The u64 value extracted from data_decoded.

alternatively, the data could be flattened to each metric key/value are columns and the table grain would be at the transaction level. This will work better for reporting because we would not need to pivot the table

**Requirements:**
- Incremental Refresh: The table should be modeled in DBT and updated incrementally to minimize processing costs.
- Validation: The sum of metric_value in this new table must match the raw event logs.
- Full Soroban history: table must be fully refreshed to the launch of Soroban (Feb 2024)

### What alternatives are there?

1. **BigQuery Search Indexes:** We could add a search index to the raw JSON column. While this helps with filtering, it does not solve the compute-heavy cost of the mathematical aggregations (SUM/AVG) needed for utilization metrics.

2. **Materialized Views:** We could create a Materialized View on top of the raw table. However, since the source table is heavily partitioned and clustered, a standard DBT incremental model offers more control over the backfill and schema evolution.

3. **Status Quo:** Continuing to query raw events, which will eventually lead to dashboard timeouts as network activity increases throughout 2026.

PoojaDoctor:
## Summary

- Adds `soroban_core_metrics` incremental model that pre-parses `core_metrics` diagnostic contract events, enabling efficient Soroban network utilization queries without full table scans and JSON parsing
- Adds `config_settings_snapshot` table model with SCD2 history of network config settings, enriched with metric mapping seed for utilization analysis
- Adds `config_setting_to_metrics_mapping` seed mapping config setting names to core metrics, grain, and utilization relevance

Closes #189

## Test plan

- [ ] Verify `soroban_core_metrics` incremental logic filters correctly by `batch_start_date`/`batch_end_date` with partition pruning
- [ ] Verify `config_settings_snapshot` correctly unpivots all config settings and applies SCD2 logic
- [ ] Verify seed join enriches config settings with `core_metric`, `grain`, `is_resource_utilization_relevent`, `pretty_config_name`
- [ ] Run `dbt test` to confirm uniqueness, not_null, and accepted_values constraints pass
- [ ] Validate output matches reference utilization query against production tables

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -82,6 +82,8 @@ models:
 
 seeds:
   stellar_dbt_public:
+    config_setting_to_metrics_mapping:
+      +dataset: crypto_stellar_dbt
     dbt_project_evaluator_config:
       +dataset: dbt_project_evaluator
   dbt_project_evaluator:
--- a/models/docs/marts/soroban/config_settings_snapshot.md
+++ b/models/docs/marts/soroban/config_settings_snapshot.md
@@ -1,4 +1,4 @@
-[comment]: < Configuration Settings Snapshot -
+[comment]: < Configuration Settings Snapshot - >
 
 {% docs config_settings_snapshot %}
 
--- a/models/docs/marts/soroban/config_settings_snapshot.md
+++ b/models/docs/marts/soroban/config_settings_snapshot.md
@@ -1,0 +1,29 @@
+[comment]: < Configuration Settings Snapshot -
+
+{% docs config_settings_snapshot %}
+
+This table captures historical snapshots of various configuration settings from the Stellar network. It unpivots different metric values from the `config_settings` table and tracks their validity over time.
+
+Each metric represents a specific setting related to transaction limits, ledger constraints, fees, and other network parameters. The `valid_from` and `valid_to` timestamps help in identifying when each setting was active.
+
+The table is enriched with metric mapping data from the `config_setting_to_metrics_mapping` seed, providing core metric names, granularity levels, and human-readable display names for each config setting.
+
+{% enddocs %}
+
+{% docs config_setting_name %}
+
+The name of the configuration setting (metric) extracted from the source table. Each row represents a different setting tracked over time.
+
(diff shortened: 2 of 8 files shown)


PoojaDoctor:
Close for now.

PR #225, opened 2026-04-08
Refactor generic tests to use batch start and end dates

sydneynotthecity:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What
Migrated all generic and singular tests from the `dbt_airflow_macros` package to use `batch_start_date` / `batch_end_date` variables, allowing for more targeted testing during backfilling. Also optimized the uniqueness tests for BigQuery performance on high-cardinality tables. The refactor supports improving the entity attribution pipeline runtime, which is now consistent 75+ minutes for a single day.

Specifically:
 - Refactored all 5 generic tests (incremental_accepted_values, incremental_not_null, incremental_unique, incremental_unique_combination_of_columns, incremental_expression_is_true) to use var("batch_start_date") / var("batch_end_date") instead of `dbt_airflow_macros.ts(timezone=none)`
- Refactored all 5 remaining singular tests to replace `dbt_airflow_macros.ts(timezone=none) `with var("batch_end_date") as the anchor timestamp
- Removed yu-iskw/dbt_airflow_macros from [packages.yml](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) 
- Dropped the `less_than_equal_to` parameter from all generic tests (was unused by every caller)
- Retained `greater_than_equal_to` as a deprecated fallback that widens the lower bound: [batch_start_date - interval, batch_end_date)
- Default generic test window is now [batch_start_date, batch_end_date) with an exclusive upper bound, matching model conventions
- Optimized incremental_unique and incremental_unique_combination_of_columns by replacing GROUP BY / HAVING count(*) > 1 with QUALIFY ROW_NUMBER() OVER (PARTITION BY ...) > 1 to avoid full hash aggregation on high-cardinality columns

### Why

Models and macros were fully migrated to `batch_start_date` / `batch_end_date` in a prior refactor, but all tests still depended on the `dbt_airflow_macros` package. The old pattern anchored on a single Airflow timestamp with hardcoded intervals, which was brittle and didn't align with flexible date range backfills. Tests should validate exactly the window that was loaded.

The GROUP BY / HAVING pattern in uniqueness tests is expensive on BigQuery for tables with 15-30M daily rows. QUALIFY ROW_NUMBER() avoids this by streaming through partitions without aggregation. This will speed up the entity attribution models by several minutes where we're seeing a bottleneck due to long running tests (7+ min to finish)

The anomaly detection tests were not measuring anything actionable and have been removed in favor of Datafold tests. The rest of the singular tests can be removed after confirming they run properly in Datafold.

### Known limitations

This does not refactor any specific model configurations that use these tests. We will need to audit `greater_than_or_equal_to` usage and remove/adjust as necessary.

This also does not refactor any references to `airflow_start_timestamp` which is another env var passed from airflow at runtime. That can be done at a future date


tests/eho_by_ops.sql
```diff
@@ -20,18 +20,18 @@ WITH find_missing AS (
   FROM {{ ref('stg_history_operations') }} op
   LEFT OUTER JOIN {{ ref('enriched_history_operations') }} eho
     ON op.op_id = eho.op_id
-    AND eho.closed_at >= TIMESTAMP_SUB('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL 1 DAY )
+    AND eho.closed_at >= TIMESTAMP('{{ var("batch_start_date") }}')
```
ttboogie: The eho-side and the op-side have different interval periods, which I think will cause false positives.

- `op.batch_run_date` is filtered to 1 day of data
- `eho.closed_at` is filtered to the current 30 min wide batch (Dag runs every 30 min)

Any row from the op-side that is outside of the 30 minute **eho** window (but inside the 1 day **op** window) will have a resulting NULL from the left join.

Should filter eho to be 1 day window so they align.

tests/bucketlist_db_size_check.sql
```diff
@@ -19,7 +19,7 @@ with bucket_max_size as (
     closed_at,
     total_byte_size_of_bucket_list / 1000000000 as bl_db_gb
   from {{ ref('stg_history_ledgers') }}
-  where closed_at >= TIMESTAMP_SUB('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL 1 HOUR )
+  where closed_at >= TIMESTAMP('{{ var("batch_start_date") }}')
```
ttboogie: Heads up: this changes the lookback from 1 hour to 30 minutes, since `dbt_singular_tests_dag` runs */30.

tests/sorobon_surge_pricing_check.sql
```diff
@@ -14,7 +14,8 @@ with surge_pricing_check as (
     ledger_sequence,
     closed_at
 from {{ ref('enriched_history_operations_soroban') }}
-where closed_at >= TIMESTAMP_SUB('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL 1 HOUR )
+where closed_at >= TIMESTAMP('{{ var("batch_start_date") }}')
```
ttboogie: Same as `bucketlist_db_size_check.sql`: lookback shrank from 1 hour to 30 min (the DAG runs every 30 minutes). Plus a new upper bound that didn't exist before.

tests/num_txns_and_ops.sql
```diff
@@ -18,7 +18,7 @@ WITH txn_count AS (
     FROM {{ ref('stg_history_transactions') }}
     --Take all ledgers committed in the last 36 hours to validate newly written data
     -- Alert runs at 12pm UTC in GCP which creates the 36 hour interval
-    WHERE batch_run_date >= DATETIME(TIMESTAMP_SUB('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL 1 DAY ))
+    WHERE batch_run_date >= DATETIME(TIMESTAMP('{{ var("batch_start_date") }}'))
```
ttboogie: Bigger shrink here: was 24 hours of `batch_run_date`, now 30 minutes (the DAG runs */30). Same change at lines 30, 31, 47.

If intentional, the comment block at lines 19-20 is now misleading: it still describes a 36-hour window.

tests/num_txns_and_ops.sql
```diff
@@ -63,6 +63,6 @@ SELECT batch_id,
 FROM raw_values
 --@TODO: figure out a more precise delay for ledgers. Since tables are loaded on a 15-30 min delay,
 -- we do not want a premature alert to row count mismatches when it could be loading latency
-WHERE closed_at <= TIMESTAMP_ADD('{{ dbt_airflow_macros.ts(timezone=none) }}', INTERVAL -180 MINUTE )
+WHERE closed_at < TIMESTAMP('{{ var("batch_end_date") }}')
```
ttboogie: The old `<= ts - 180 MINUTE` was a 3-hour grace period for ledgers that hadn't finished loading yet. 

Replacing it with `< batch_end_date` removes that buffer, so I think that this test could now alert mid-load if it runs before the latest ledgers finish being written

ttboogie (approved):
Please see my comment about the left join in the `eho_by_ops.sql`.

If the interval window changes are intentional, the only other thing needed is to update a few comments within the code to match the new interval periods.

tests/generic/incremental_unique.sql
```diff
@@ -1,59 +1,43 @@
...
-          unique_field,
-          count(*) as n_records
+      {% if condition != '' %}
+      and {{ condition }}
+      {% endif %}
 
-      from dbt_test__target
-      group by unique_field
-      having count(*) > 1
+      qualify row_number() over (partition by {{ column_name }}) > 1
```
ttboogie, line 41: **Just a Note**: switching from `having count(*) > 1` to `qualify row_number() > 1` changes the output shape of the test query.  old output was one row per duplicate key with an n_records count; new output is N-1 rows per duplicate key with no count column. 

Pass/fail behavior is the same, but if we're saving these results anywhere the row counts and schema will change.

sydneynotthecity: Nice catch, in this case, we use the test just to flag duplicates, so it's ok if the result set changes slightly. It is super expensive to calculate the count on high-cardinality, large datasets, so switching to row level output is cheaper to run. If an engineer needs the count of duplicates, they'll need to calculate themselves during triaging which i think is an ok tradeoff to decrease the 5-6 min query runtime 😬

tests/bucketlist_db_size_check.sql
sydneynotthecity: yeah, that's fair. i'll add the interval look back back to the code, though i think it's overkill for this check given the upstream table `stg_history_ledgers`

tests/sorobon_surge_pricing_check.sql
sydneynotthecity: Since the test runs every /*30 min and the batch_end_date is the upper bound, this filter should be a no-op. It's just helpful if we ever ran this test for an older period of time. The test wouldn't scan full history, it would only scan the slice we wanted. I left the upper bound in there, and re-added the 1 hour lookback

tests/num_txns_and_ops.sql
sydneynotthecity: yeah, that's fair. I think 3 hours is too big of a grace period. We will be slow to be notified in the event of a real issue, so I've tightened that to 90 minutes. if it's too noisy, we can bump back out.

tests/eho_by_ops.sql
sydneynotthecity: Ha, great catch 😉 

I updated the bug. Will post my local test results as a comment once I push these changes.

sydneynotthecity:
@ttboogie I've incorporated your feedback into this PR if you can re-review

Here are also the results from my local testing 😉 
```
(stellar_dbt_public) (base) szwise@szwise-C02FW0SMQ05P-mbp stellar-dbt-public % source .env && IS_SINGULAR_AIRFLOW_TASK="true" BATCH_START_DATE="2026-05-08 14:00:00+00:00" BATCH_END_DATE="2026-05-08 15:00:00+00:00" dbt test --select tag:singular_test

15:23:05  Running with dbt=1.11.0-rc3
/Users/szwise/Code/stellar-dbt-public/stellar_dbt_public/lib/python3.13/site-packages/google/cloud/aiplatform/models.py:52: FutureWarning: Support for google-cloud-storage < 3.0.0 will be removed in a future version of google-cloud-aiplatform. Please upgrade to google-cloud-storage >= 3.0.0.
  from google.cloud.aiplatform.utils import gcs_utils
15:23:08  Registered adapter: bigquery=1.10.3
15:23:09  Unable to do partial parsing because change detected to override macro. Starting full parse.
15:23:13  [WARNING][CustomKeyInConfigDeprecation]: Deprecated functionality
Custom key `snapshot_end_date` found in `config` in file
`models/snapshots/liquidity_pools_snapshot.sql`. Custom config keys should move
into the `config.meta`.
15:23:15  [WARNING]: Found patch for macro "incremental_snapshot" which was not found
15:23:15  [WARNING][MissingArgumentsPropertyInGenericTestDeprecation]: Deprecated
functionality
Found top-level arguments to test `incremental_not_null` defined on
'stg_accounts' in package 'stellar_dbt_public'
(models/staging/stg_accounts.yml). Arguments to generic tests should be nested
under the `arguments` property.
15:23:24  Found 98 models, 1267 data tests, 1 seed, 2 operations, 19 sources, 1535 macros
15:23:24  
15:23:24  Concurrency: 1 threads (target='prod')
15:23:24  
15:23:28  Elementary: Runtime data: {"config": {}, "dbt_version": "1.11.0rc3", "elementary_version": "0.21.0", "database": "crypto-stellar", "schema": "elementary"}
15:23:28  1 of 1 START hook: elementary.on-run-start.0 ................................... [RUN]
15:23:28  1 of 1 OK hook: elementary.on-run-start.0 ...................................... [OK in 0.38s]
15:23:28  
15:23:28  1 of 5 START test bucketlist_db_size_check ..................................... [RUN]
15:23:28  [WARNING]: Installed package 'elementary' is overriding the built-in
materialization 'test'. Overrides of built-in materializations from installed
packages will be deprecated in future versions of dbt. For more information:
https://docs.getdbt.com/reference/global-configs/legacy-behaviors
15:23:30  1 of 5 PASS bucketlist_db_size_check ........................................... [PASS in 2.38s]
15:23:30  2 of 5 START test eho_by_ops ................................................... [RUN]
15:23:34  2 of 5 PASS eho_by_ops ......................................................... [PASS in 3.17s]
15:23:34  3 of 5 START test ledger_sequence_increment .................................... [RUN]
15:23:35  3 of 5 PASS ledger_sequence_increment .......................................... [PASS in 1.60s]
15:23:35  4 of 5 START test num_txns_and_ops ............................................. [RUN]
15:23:38  4 of 5 PASS num_txns_and_ops ................................................... [PASS in 2.60s]
15:23:38  5 of 5 START test sorobon_surge_pricing_check .................................. [RUN]
15:23:41  5 of 5 WARN 168572 sorobon_surge_pricing_check ................................. [WARN 168572 in 2.92s]
15:23:41  
15:23:41  Missing Elementary models in 'crypto-stellar.elementary'. Please run 'dbt run -s elementary --target prod'.
15:23:41  Missing Elementary models in 'crypto-stellar.elementary'. Please run 'dbt run -s elementary --target prod'.
15:23:41  1 of 1 START hook: elementary.on-run-end.0 ..................................... [RUN]
15:23:41  1 of 1 OK hook: elementary.on-run-end.0 ........................................ [OK in 0.50s]
15:23:41  
15:23:41  Finished running 2 project hooks, 5 data tests in 0 hours 0 minutes and 16.87 seconds (16.87s).
15:23:42  
15:23:42  Completed with 1 warning:
15:23:42  
15:23:42  Warning in test sorobon_surge_pricing_check (tests/sorobon_surge_pricing_check.sql)
15:23:42  Got 168572 results, configured to warn if != 0
15:23:42  
15:23:42    compiled code at target/compiled/stellar_dbt_public/tests/sorobon_surge_pricing_check.sql
15:23:42  
15:23:42  Done. PASS=6 WARN=1 ERROR=0 SKIP=0 NO-OP=0 TOTAL=7
```

I had to do quite a bit of manuevering to get these to run locally. If you want a spot-check or rundown on what i did, lmk

ttboogie (approved):
lgtm

PR #248, opened 2026-05-06
contract token updates

hunterpack:
<details>

### PR Structure

- [x] This PR has reasonably narrow scope (if not, break it down into smaller PRs).
- [x] This PR avoids mixing refactoring changes with feature changes (split into two PRs otherwise).

### Thoroughness

- [x] This PR adds tests for the most critical parts of the new functionality or fixes.
- [x] I've updated the docs and README with the added features, breaking changes, new instructions on how to use the repository.

### Release planning

</details>

### What

Adds a new `int_contract_asset_codes` intermediate model that resolves a non-null `asset_code` for every contract by coalescing, in priority order:

1. `asset_code` from SAC token transfer events (`stg_assets`),
2. SEP-41 `symbol` from contract storage metadata (`int_asset_metadata`), then
3. the `contract_id` itself as a final fallback.

The new model is consumed by:

- `int_account_balances__contracts` — flowing into `asset_balances__daily_agg` and `account_balances__daily_agg`.
- `int_token_transfer_enrichment` — flowing into the public `token_transfers` mart (and from there into stellar-dbt's `int_account_activity__smart_contracts` once that repo's `packages.yml` is bumped).

The resolver also exposes an `asset_code_source` enum (`sac` | `metadata` | `contract_id`) so analysts can distinguish trusted asset_codes from fallback values.

### Why

Closes [stellar/stellar-dbt#677](https://github.com/stellar/stellar-dbt/issues/677). Contract tokens were reporting `null` for `asset_code` in these marts, which caused incorrect aggregations in Omni dashboards and effectively hid custom-contract-token activity in reports.

### Validation against production data (2026-05-06)

| Metric | Count |
|---|---|
| Total contracts known to the network | 403,342 |
| With SAC asset_code | 384,595 (95.4%) |
| With SEP-41 symbol metadata | 5,975 (1.5%) |
| Have both | 2,098 |
| SAC-only | 382,497 |
| Metadata-only | 3,877 |
| Neither (fall back to contract_id) | 14,870 (3.7%) |
| **SAC asset_code ↔ symbol mismatches** | **1** |

The single mismatch is the native XLM SAC (`CAS3J7GYLGXMF6TDJBBYYSE3HQ6BBSMLNUQ34T6TZMYMW2EVH34XOWMA`), which publishes `symbol = "native"` while staging normalizes the asset_code to `"XLM"`. Confirmed as desired behavior — prioritizing SAC asset_code over symbol is correct. The new singular test excludes this known case and warns on any future mismatches.

### Tests added

- `int_contract_asset_codes.yml`: `unique` + `not_null` on `contract_id`, `not_null` on `asset_code`, `accepted_values` on `asset_code_source`. The `not_null` on `asset_code` is the direct regression test for the coalesce logic — it fires on every test run regardless of date scope.
- `int_account_balances__contracts.yml`: `relationships` test ensuring every `contract_id` resolves via `int_contract_asset_codes`. Validates the join logic end-to-end, not just stale data.
- `tests/sac_asset_code_matches_metadata_symbol.sql`: singular warn-level test for SAC asset_code / SEP-41 symbol mismatches (with the known XLM exception).

### Known limitations

- **Backfill required.** `asset_balances__daily_agg`, `account_balances__daily_agg`, and `token_transfers` are incremental — historical rows with `null` asset_code will persist until those marts are re-run with `--full-refresh` (or the relevant date ranges backfilled). Tracked separately.
- **Symbol coverage is sparse** (1.5% of contracts publish SEP-41 `symbol`), so ~3.7% of contracts still surface their `contract_id` in the `asset_code` column. This is the intended fallback — it groups custom-contract activity per-contract rather than under `null`.

### Testing
- [ ] I have tested my changes locally or in dev airflow
- [ ] I have added or updated relevant datafold alerts


models/intermediate/int_contract_asset_codes.sql
```diff
@@ -0,0 +1,55 @@
...
+
+    , all_contracts as (
+        select contract_id from asset_per_contract
+        union distinct
+        select contract_id from metadata
+    )
+
+select
+    c.contract_id
+    , coalesce(a.asset_code, m.symbol, c.contract_id) as asset_code
```
hunterpack: falls back to the `contract_id` itself so that the column is never null. This satisfies the issue's "no null asset_codes for contract tokens" requirement and keeps Omni aggregations from bucketing custom tokens together under a single null group; per the 2026-05-06 validation, 3.7% of contracts (14,870 rows) actually hit this fallback, meaning a 56-character C… address ends up sitting in a column otherwise holding 4–12 char tickers — should we keep it (asset_code is guaranteed non-null and per-contract groupable, but the column's value space is now mixed) or drop it (asset_code stays a "real" ticker, but custom contract tokens with no SEP-41 metadata go back to surfacing as null and the original bug partially returns for that 3.7%)?

models/intermediate/int_asset_metadata.sql
```diff
@@ -0,0 +1,55 @@
...
+        select contract_id from asset_per_contract
+        union distinct
+        select contract_id from metadata
+    )
+
+select
+    c.contract_id
+    , coalesce(a.asset_code, m.symbol, c.contract_id) as asset_code
+    , a.asset_issuer
+    , a.asset_type
```
hunterpack, line 82: `asset_type` here is just passed through from stg_assets and is empty-string/null for custom contract tokens (everything that isn't a classic SAC). That produces the same null-aggregation problem in Omni that motivated the asset_code fix — should we add coalesce(nullif(a.asset_type, ''), 'contract') as asset_type here so contract-token rows bucket under a real value? Cheap to add (one line), but it changes the public mart's asset_type value space from {'native', 'credit_alphanum4', 'credit_alphanum12'} to also include 'contract', so I want to confirm before folding it in.

models/intermediate/int_contract_asset_codes.sql
```diff
@@ -0,0 +1,55 @@
...
+with
+    asset_per_contract as (
+        select
+            asset_contract_id as contract_id
+            , max(case when asset_code is not null and asset_code != '' then asset_code end) as asset_code
+            , max(case when asset_issuer is not null and asset_issuer != '' then asset_issuer end) as asset_issuer
+            , max(case when asset_type is not null and asset_type != '' then asset_type end) as asset_type
+        from {{ ref('stg_assets') }}
+        where asset_contract_id is not null
+        group by 1
```
amishas157: `asset_contract_id` should be unique according to 
https://github.com/stellar/stellar-dbt-public/blob/50b873cd48763fef747bf76c671683f2eb9b2def/models/staging/stg_assets.yml#L17-L21

group by may not be needed

models/intermediate/int_contract_asset_codes.sql
amishas157: Would like @PoojaDoctor to confirm this. If she wants to coalesce(asset_code, contract_id) downstream or we set it here.

models/intermediate/tokens/transfers/int_token_transfer_enrichment.sql
```diff
@@ -41,9 +43,11 @@ with
         from {{ ref('stg_token_transfers_raw') }} as tt
         left join {{ ref('int_asset_metadata') }} as metadata
             on tt.contract_id = metadata.contract_id
+        left join {{ ref('int_contract_asset_codes') }} as ac
```
amishas157: It looks int_contract_asset_codes is superset of int_asset_metadata. We should probably replace int_asset_metadata by int_contract_asset_codes or update int_asset_metadata to contain logic of asset codes and remove int_contract_asset_codes

models/intermediate/account_balances/int_account_balances__contracts.sql
```diff
@@ -135,5 +135,5 @@ select
     , a.asset_code
     , agg.balance
 from agg
-left join {{ ref('stg_assets') }} as a
-    on agg.contract_id = a.asset_contract_id
+left join {{ ref('int_contract_asset_codes') }} as a
```
amishas157: 👍

models/intermediate/int_contract_asset_codes.sql
hunterpack: fixed

models/intermediate/int_contract_asset_codes.sql
PoojaDoctor: up to you, as long as the final tables (public, gold, omni_pdts) are correct, I care a lot less about where this sits.

PoojaDoctor: we won't be using int__

models/intermediate/account_balances/int_account_balances__contracts.sql
```diff
@@ -30,7 +30,7 @@ with
             -- Note that 10^-7 is the default precision for Stellar assets. Recognized assets with default precision will have null token_precision.
             , sum(safe_cast(tt.amount_raw as numeric) * pow(10, coalesce(-safe_cast(metadata.decimal as int64), -7))) as balance
         from {{ ref('stg_token_transfers_raw') }} as tt
```
PoojaDoctor: im curious why not use your enriched token transfers here, instead of raw?

models/intermediate/tokens/transfers/int_token_transfer_enrichment.sql
```diff
@@ -28,22 +28,24 @@ with
...
+            , sum(safe_cast(tt.amount_raw as numeric) * pow(10, coalesce(-safe_cast(ac.decimal as int64), -7))) as amount
             , tt.contract_id
             , tt.ledger_sequence
             , tt.closed_at
             , tt.to_muxed
             , tt.to_muxed_id
         from {{ ref('stg_token_transfers_raw') }} as tt
-        left join {{ ref('int_asset_metadata') }} as metadata
-            on tt.contract_id = metadata.contract_id
+        left join {{ ref('int_contract_asset_codes') }} as ac
```
PoojaDoctor: how are you guys building the marts version downstream? should you switch sources?

models/intermediate/int_asset_metadata.sql
```diff
@@ -1,48 +0,0 @@
-{% set meta_config = {
-    "tags": ["assets"]
-} %}
```
PoojaDoctor, line 3: my nit is on naming, int_asset_metadata is more intuitive than the int_contract_asset_codes if the latter now has more info

models/intermediate/int_contract_asset_codes.sql
```diff
@@ -0,0 +1,93 @@
...
+
+    , all_contracts as (
+        select contract_id from asset_per_contract
+        union distinct
+        select contract_id from metadata
+    )
+
+select
+    c.contract_id
+    , coalesce(a.asset_code, m.symbol, c.contract_id) as asset_code
```
PoojaDoctor: have you checked if recognized asset codes are changing to contract id? I'm not sure how you want to deal with this in the public repo, but might want to make changes in the private one.

PoojaDoctor (changes requested):
left some comments.

models/intermediate/int_contract_asset_codes.sql
PoojaDoctor: sorry correction. I though this was just isolated to the int__
no I think asset code should be blank if we don't have one, unless its a recognized asset. we shouldn't surface contract id as the asset code

sydneynotthecity: > no I think asset code should be blank if we don't have one, unless its a recognized asset


why wouldn't we want to set the symbol if a non-recognized contract token has a short code/symbol? 
I agree that i don't like the contract id as a fallback, but i think if we have access to metadata.symbol for another contract token, that should be surfaced
otherwise this will be yet another table to backfill if/when assets change on the recognized asset seed file

models/intermediate/account_balances/int_account_balances__contracts.sql
hunterpack: Three reasons for staying on stg_token_transfers_raw:

1. Only the raw columns are needed. This model uses from, to, contract_id, amount_raw, asset_type, and closed_at, all available pre-enrichment. The columns int_token_transfer_enrichment adds (event_type, is_soroban, unique_key, the SHA hash) would be unused here.
2. Materialization mismatch. int_token_transfer_enrichment is incremental, partitioned on closed_at. This model intentionally rebuilds from 2024-02-20 to current on every run (see the header comment) because the contract-balance volume is small. Pulling from an incremental upstream would force this model to either inherit those partitions or coordinate full-refreshes.
3. `int_asset_metadata` is the actual shared piece. Both models join `int_asset_metadata` for the decimal scaling; that's where the de-duplication lives. The raw view is the minimum required dependency for this model's per-day-per-account grain.

models/intermediate/tokens/transfers/int_token_transfer_enrichment.sql
hunterpack: The downstream mart (token_transfers in marts/tokens/transfers/) already reads from `int_token_transfer_enrichment`, so the `asset_code` coalesce we added here propagates automatically, the mart's `asset_code` column will now be coalesce(raw asset_code, SEP-41 symbol) instead of just raw `asset_code`, with null when neither is available. No source switch is needed, the fix lands in the public mart without any further changes downstream.

models/intermediate/int_contract_asset_codes.sql
hunterpack: @sydneynotthecity agreed, that's how this is working. We kept `coalesce(a.asset_code, m.symbol)`, only the third-tier `contract_id` fallback got dropped per Pooja's preference. Non-recognized contract tokens that publish a SEP-41 symbol will still surface that symbol, only contracts publishing neither a SAC `asset_code` nor SEP-41 metadata go to null. That preserves the "minimize backfill when recognized_assets changes" property you flagged.

models/intermediate/int_asset_metadata.sql
hunterpack: updated to your preference

models/intermediate/int_contract_asset_codes.sql
hunterpack: No longer applicable after the slack conversation + the latest commits.

models/intermediate/int_contract_asset_codes.sql
PoojaDoctor: yup hunter, thats exactly right. ty for making the change!

models/intermediate/account_balances/int_account_balances__contracts.sql
PoojaDoctor: got it. my POV was cleaner lineage, raw -> intermediate transfers -> other intermediate tables building on top. if this table is a full refresh would it be impacted if the underlying is partitioned. on 3, my thinking is it prevents us from joining the same table twice (easier to maintain)

models/intermediate/account_balances/int_account_balances__contracts.sql
```diff
@@ -135,5 +135,5 @@ select
     , a.asset_code
     , agg.balance
 from agg
-left join {{ ref('stg_assets') }} as a
-    on agg.contract_id = a.asset_contract_id
+left join {{ ref('int_asset_metadata') }} as a
```
amishas157, line 133: [nit] There is already a join with int_asset_metadata in `token_transfers_from` and `token_transfers_to` clause above. So asset info could be fetched in existing joins.

tests/sac_asset_code_matches_metadata_symbol.sql
```diff
@@ -0,0 +1,19 @@
+-- For contract tokens that publish both an asset_code via SAC transfer events
+-- and a SEP-41 symbol in their contract storage metadata, the two should agree.
+-- A mismatch indicates either a misconfigured SAC, a custom contract impersonating
+-- a recognized asset, or a data-quality issue worth investigating.
+--
+-- Known exception: the XLM SAC contract publishes "native" as its symbol while
+-- our staging layer rewrites the asset_code to "XLM" for ergonomics.
+-- This was confirmed against production data on 2026-05-06; no other mismatches existed.
+{{ config(severity='warn') }}
```
amishas157: It will also need tag and enabled clause like https://github.com/stellar/stellar-dbt-public/blob/c3e59d39ba128e3ace900f106821c2316983105e/tests/eho_by_ops.sql#L3-L10 to be run in dbt.

Also, should we just set this up in datafold instead?

amishas157 (approved):
Left a small nit inline. Otherwise looks good to me.
Note that you would need to add config in test so that it runs in singular test dag.

models/intermediate/account_balances/int_account_balances__contracts.sql
hunterpack: updated

models/intermediate/account_balances/int_account_balances__contracts.sql
hunterpack: updated

tests/sac_asset_code_matches_metadata_symbol.sql
hunterpack: added, also agree we should look at list of tests to set up in datafold

models/intermediate/tokens/transfers/int_token_transfer_enrichment.sql
```diff
@@ -28,22 +28,25 @@ with
             -- Original token transfers only built asset_code:asset_issuer, it dropped asset_type for all assets less XLM
             , replace(tt.asset, substr(tt.asset, 1, instr(tt.asset, ':')), '') as asset
             , tt.asset_type
-            , tt.asset_code
+            -- For contract tokens that don't expose asset_code on transfer events, fall back to
+            -- int_asset_metadata which coalesces with SEP-41 symbol metadata (null when neither
+            -- a SAC asset_code nor a SEP-41 symbol is available).
+            , coalesce(nullif(tt.asset_code, ''), ac.asset_code) as asset_code
```
hunterpack, line 34: fix

PR #249, opened 2026-05-06
Harden CI Linting and Release Drafter workflows

amishas157:
## Summary

- **diff-quality.yml**: bind `github.head_ref` to a step `env:` var and quote it in the shell. Removes the script-injection surface from a crafted PR branch name (matches GitHub's [Actions hardening guide][1]).
- **release.yml**: migrate the deprecated `::set-output` workflow command to `$GITHUB_OUTPUT`, and pass branch / version values through `env:` instead of re-interpolating them into bash. Closes the same injection class on the release path and unblocks future runner-image upgrades that drop `::set-output` support.

Impact under the current configuration is limited (fork PRs run with read-only `GITHUB_TOKEN` and no secrets, and the release job only runs on merged PRs which require write access), but the fixes are mechanical defense-in-depth and remove a tripwire if the workflow is ever reordered or the trigger changes.

## Test plan

- [ ] CI Linting workflow runs successfully on this PR (exercises the changed `diff-quality.yml` step).
- [ ] After merge, the next merged PR triggers Release Drafter and tags + publishes a release as before.

[1]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,37 +29,42 @@ jobs:
 
       - name: Get branch name
         id: getbranch
-        run: echo ::set-output name=BRANCH::${GITHUB_HEAD_REF}
+        env:
+          HEAD_REF: ${{ github.head_ref }}
+        run: echo "BRANCH=$HEAD_REF" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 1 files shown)


hunterpack (approved):
lgtm

PR #256, opened 2026-05-14
Detect stale latest closed_at in no_missing_ledgers

amishas157:
## Summary
- The existing `no_missing_ledgers` singular test only finds gaps *between existing rows*, so a table that stops receiving data silently passes.
- Adds a `stale_latest` CTE that fails when `max(closed_at)` is `null` (no rows in the 7-day lookback) or `< batch_end_date` (latest data hasn't caught up).
- Unifies results into a `failures` CTE with a `failure_reason` column (`'gap'` or `'stale_latest'`) so the test output makes the cause obvious. Severity stays `error` for both.

## Test plan
- [x] Trigger the airflow singular-test task on a batch where upstream data is current — expect 0 rows.
- [x] Manually run with a `batch_end_date` set ahead of `max(closed_at)` in one of the listed tables — expect a `stale_latest` row for that table.
- [x] Confirm existing gap-detection path still flags engineered sequence gaps (`failure_reason = 'gap'`).
- [x] Confirm `table_group_name` / `table_name` var scoping still narrows both checks.


tests/no_missing_ledgers.sql
```diff
@@ -107,5 +95,43 @@ with all_gaps as (
     {% endif %}
 )
 
+, stale_latest as (
```
amishas157, line 98: This test ensures that data until batch_end_date(with 5 minutes lag threshold) is available in the table.
The downstream task checks for upstream tables.

chowbao (approved):

PR #267, opened 2026-06-05
Asset code contract tokens v2

hunterpack:
### What

Restores the HUBBLE-677 contract token resolution work (originally PR #248, reverted by #262), plus two new safeguards on top:

1. **SAC sibling decimals fallback** in `int_asset_metadata`: when a SAC and a SEP-41 contract share the same `asset_code` AND the SAC's `asset_issuer` matches the SEP-41's `admin`, the SAC row inherits the SEP-41's `decimal` instead of returning null.
2. **New singular test** `int_token_transfer_enrichment_amount_matches_current_metadata` that recomputes stored `amount` against current `int_asset_metadata.decimal` and fails on rows whose stored value disagrees by more than 100 absolute units.

### Why

#248 was reverted on 2026-05-18 because pre-existing stale rows from January 2026 in `int_token_transfer_enrichment` (scaled at 10⁻⁷ instead of 10⁻¹⁸ due to a metadata-resolution gap at write time) got surfaced by the refactored `int_account_balances__contracts`. The deJAAA/deJTRSY balances on Margarita's Omni RWA chart inflated by ~10¹¹x.

The two new safeguards prevent recurrence:
- The sibling fallback ensures SACs of higher-decimal SEP-41 tokens get the correct decimals in `int_asset_metadata`, so downstream amount scaling doesn't default to 10⁻⁷.
- The singular test catches any stale-baked-in scaling in enrichment, regardless of when it was introduced. Catches the original issue and any future metadata-resolution gaps.

Full context in stellar-dbt's `POSTMORTEM_DEJAAA_DEJTRSY_2026-05-15.md`.

### Known limitations

- The SAC sibling fallback requires `asset_issuer = admin` to match. For tokens where the SEP-41's `admin` is a contract address (C-address) rather than the issuer's G-address, the fallback won't trigger. Acceptable for the deJAAA/deJTRSY case which motivated this change.
- The singular test tolerance is set to 100 absolute units; tight enough to catch genuine scaling regressions (typically off by 10¹¹+) but loose enough to ignore float64 precision noise on huge summed values.

### Testing
- [x] I have tested my changes locally or in dev airflow
- [ ] I have added or updated relevant datafold alerts

### DIFF
--- a/models/intermediate/int_asset_metadata.sql
+++ b/models/intermediate/int_asset_metadata.sql
@@ -9,12 +9,26 @@
 }}
 
 with
-    unique_contracts as (
+    asset_per_contract as (
+        -- stg_assets.asset_contract_id is unique (see stg_assets.yml), so no aggregation needed.
+        select
+            asset_contract_id as contract_id
+            , nullif(asset_code, '') as asset_code
+            , nullif(asset_issuer, '') as asset_issuer
+            , nullif(asset_type, '') as asset_type
+        from {{ ref('stg_assets') }}
+        where asset_contract_id is not null
+    )
+
+    -- SEP-41 / SAC metadata read directly from contract storage.
+    , unique_contracts as (
         select distinct contract_id
         from {{ ref('stg_token_transfers_raw') }}
         -- filter only for soroban contracts
         where closed_at >= '2024-02-01'
-    ), contract_metadata as (
+    )
+
+    , contract_metadata as (
         select
             uc.contract_id
             , cd.val_decoded
@@ -22,7 +36,9 @@ with
         left join {{ ref('contract_data_current') }} as cd
             on uc.contract_id = cd.contract_id
         where cd.contract_key_type = 'ScValTypeScvLedgerKeyContractInstance'
-    ), flattened_data as (
+    )
+
+    , flattened_data as (
         select
             contract_id
             , json_value(m.key.symbol) as metadata_key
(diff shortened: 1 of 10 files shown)


models/intermediate/int_asset_metadata.sql
```diff
@@ -36,13 +52,68 @@ with
...
+        group by contract_id
+    )
+
+    , all_contracts as (
+        select contract_id from asset_per_contract
+        union distinct
+        select contract_id from metadata
+    )
+
+    -- SAC rows carry no contract-storage metadata of their own, so `decimal` is null on
```
hunterpack, line 75: @PoojaDoctor - Added a SAC sibling decimals fallback in this PR: when a SAC row has null decimals and there's a SEP-41 contract with the same asset_code AND matching issuer/admin, the SAC inherits the SEP-41's decimals. Prevents SACs of higher-decimal SEP-41 tokens (like deJAAA/deJTRSY at 18) from defaulting to 7 in downstream amount scaling.

hunterpack:
Adding a singular test as a safeguard for the kind of stale-baked-in scaling that caused the May 15 regression. It recomputes the stored amount against current int_asset_metadata.decimal and fails if any row disagrees by more than 100 absolute units (tolerance set to ignore float64 noise on huge summed values but catch real scaling regressions which are typically off by 10^11). Runs only in prod as part of the singular_test Airflow task, same pattern as eho_by_ops and sac_asset_code_matches_metadata_symbol.

ttboogie:
Full backfill on `int_token_transfer_enrichment` is necessary because of possible preexisting fallbacks to the previous default 10^-7 decimal value.

ttboogie (approved):
Looks good to me. Please update PR description, and see my comment about the required backfill.

PR #266, opened 2026-06-04
Microbatch int_token_transfer_enrichment and token_transfers

amishas157:
## What

Migrate following models to microbatch strategy:
- int_account balances* and account_balances
- int_token_transfer_enrichment and token_transfers
- int_tvl_accounts, int_tvl_trustlines
- eho

Run times for full refreshes of following marts:
1. TVL :  ~1 hour
2. asset_balance_agg: ~2 hours
3. token_transfer: 47
4. fee stats: 35 mins
5. EHO: 2 hours

## Companion PR

stellar/stellar-dbt#994 — `+event_time` on the staging inputs and the package pin.

### Testing Omni

Screenshot 2026-07-14 at 4 25 30 PM
Screenshot 2026-07-14 at 4 25 42 PM

Screenshot 2026-07-14 at 4 29 51 PM
Screenshot 2026-07-14 at 4 30 04 PM

PR #291, opened 2026-07-10
Update stg_contract_data uniqueness test to use ledger_sequence

amishas157:
Change the `incremental_unique_combination_of_columns` test grain on `stg_contract_data` from `[ledger_key_hash, last_modified_ledger, ledger_entry_change]` to `[ledger_key_hash, ledger_sequence, ledger_entry_change]`.

Uniqueness of a contract-data row within a ledger change is keyed by the ledger it was observed in (`ledger_sequence`), not the ledger it was last modified in (`last_modified_ledger`) — the latter repeats across observed ledgers. `ledger_sequence` already exists in the model (`stg_contract_data.sql`).

Standalone change, independent of the microbatch work. Targeting `release-v20260714`.

### DIFF
--- a/models/staging/stg_contract_data.yml
+++ b/models/staging/stg_contract_data.yml
@@ -17,10 +17,10 @@ models:
           greater_than_equal_to: "2 day"
           combination_of_columns:
             - ledger_key_hash
-            - last_modified_ledger
+            - ledger_sequence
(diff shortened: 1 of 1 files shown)


amishas157:
Folding this into #266 instead.

PR #298, opened 2026-07-14
Disable claimable_balances_current model

chowbao:
## Description

Disables the `claimable_balances_current` mart via `+enabled: false` in `dbt_project.yml`.

Also removes `claimable_balances_current` from the `current_tables` list in `tests/no_missing_ledgers.sql` — with the model disabled, the `ref()` there would fail compilation when the singular test task runs in Airflow (`IS_SINGULAR_AIRFLOW_TASK=true`).

## Verification

- `dbt parse` succeeds
- `dbt ls --select claimable_balances_current` → no enabled nodes
- `IS_SINGULAR_AIRFLOW_TASK=true dbt ls --select no_missing_ledgers` still resolves
- `pre-commit` passes on changed files

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,3 +309,7 @@ In order to apply git tagging properly, the last commit message from a repo must
 # Futher Development
 
 TODO
+
+---
+
+> **Note:** This repository is not in scope for the Stellar Development Foundation bug bounty program. Vulnerabilities found in this repo are not eligible for rewards.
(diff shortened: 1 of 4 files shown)


taha-aziz (approved):
think you might need to remove these alerts after in datafold-monitors:
- `crypto_stellar_dbt/claimable_balances_current.yaml.j2`. This is a freshness monitor that once the model stops building, the table goes stale and this monitor will fire 
- `anomaly-detection/asset_balances_equal_to_stellar_events.yaml.j2`. This queries claimable_balances_current directly (from ... claimable_balances_current as cb join recognized_assets) and on frozen data it will fire too

PR #299, opened 2026-07-15
Add datadiff configs to mart models for WAP backfill validation

amishas157:
Adds a `config.datadiff` block to each mart/gold/omni_pdt model so the `dbt_backfill`DAG runs a datadiff between prod and prod_backfill datasets. (Also relevant for staging)

Each block defines:
- **`unique_key`** → Datafold `pk_columns` (from the model's config or uniqueness test)
- **`exclude_columns`** → the load/metadata columns present on the model (`batch_id`, `batch_run_date`, `batch_insert_ts`, `airflow_start_ts`) that legitimately differ live-vs-backfill
- **`min_match_percent: 98`**
- **`filters: {column, default_start}`** for time-partitioned models, so the diff is bounded to a window — `resolve_models` injects the backfill DAG's `event_time` window and falls back to `default_start`

### DIFF
--- a/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
+++ b/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
@@ -16,11 +16,17 @@
     cluster_by=["asset_type", "asset_code", "asset_issuer"],
 )}}
 
+{# model.batch is only set at microbatch run time; fall back to the batch
+   vars so parse/compile (e.g. `dbt compile` in CI) renders without a batch
+   context. #}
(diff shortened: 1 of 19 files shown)


hunterpack (approved):
lgtm, just one question. 

`min_match_percent: 98` is a reasonable default across the board however small dimension tables lets a 2% divergence pass validation, do we want to bump to 100 for any of these tables

amishas157:
> lgtm, just one question.
> 
> `min_match_percent: 98` is a reasonable default across the board however small dimension tables lets a 2% divergence pass validation, do we want to bump to 100 for any of these tables

Lets keep it 98 and iterate as we do backfills.

PR #302, opened 2026-07-20
Tune incremental strategies (merge → table / insert_overwrite / microbatch)

amishas157:
Moves incremental models off implicit/explicit `merge` toward strategies with cleaner bug-recovery ergonomics (idempotent partitions / full rebuilds) instead of manual cleanup of orphaned rows. Decisions were made per-model with the team.

## current_state (10) — stay merge, now explicit
`account_signers_current`, `accounts_current`, `claimable_balances_current`, `config_settings_current`, `contract_code_current`, `contract_data_current`, `liquidity_pools_current`, `offers_current`, `trust_lines_current`, `ttl_current` — pinned `incremental_strategy='merge'` (were on the implicit default). SCD-1 upsert; correct as-is.

## → table
- `history_assets` — tiny cumulative dimension; `asset_id` test moved from windowed `incremental_unique` to plain `unique`+`not_null`.
- `evicted_keys` — unpartitioned, so hourly merge already full-scanned each run; full rebuild is comparable cost. Added whole-table `dbt_utils.unique_combination_of_columns` on `[ledger_key_hash, closed_at]`.
- `daily_fee_stats_agg` — day-grained aggregates.

## → microbatch
All microbatch models now set `concurrent_batches=true` and `copy_partitions=true` unconditionally (were gated on `flags.FULL_REFRESH`, so they only applied during full refreshes). This lets incremental runs also use concurrent batches and BigQuery partition copies. `batch_size` sizing is unchanged.

## Notes
- All `table` models produce a correct full rebuild when non-incremental (verified: no `{{ this }}` refs outside incremental branches; `is_incremental()` blocks are additive filters or have full `{% else %}` paths).
- Other windowed `incremental_not_null` tests are retained (still function on a table, scoped to recent window).
- Couldn't run dbt/sqlfluff locally (no warehouse) — relying on CI.

Paired with stellar/stellar-dbt#1033.

### DIFF
--- a/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
+++ b/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
@@ -5,13 +5,13 @@
     incremental_strategy='microbatch',
     event_time='day',
     batch_size=batch_size,
-    concurrent_batches=flags.FULL_REFRESH,
+    concurrent_batches=true,
     begin='2021-01-01',
     partition_by={
          "field": "day"
         , "data_type": "date"
         , "granularity": "day"
-        , "copy_partitions": flags.FULL_REFRESH
+        , "copy_partitions": true
     },
     cluster_by=["asset_type", "asset_code", "asset_issuer"],
 )}}
--- a/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
+++ b/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
@@ -1,4 +1,4 @@
-{% set batch_size = 'year' if flags.FULL_REFRESH else 'day' %}
+{% set batch_size = env_var('DBT_MICROBATCH_BATCH_SIZE', '') or ('year' if flags.FULL_REFRESH else 'day') %}
 
 {{ config(
     materialized='incremental',
--- a/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
+++ b/models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
@@ -1,4 +1,4 @@
-{% set batch_size = env_var('DBT_MICROBATCH_BATCH_SIZE', '') or ('year' if flags.FULL_REFRESH else 'day') %}
+{% set batch_size = microbatch_batch_size() %}
 
 {{ config(
     materialized='incremental',
--- a/models/marts/daily_fee_stats_agg.sql
+++ b/models/marts/daily_fee_stats_agg.sql
@@ -20,11 +20,6 @@ with
     ledger_stats as (
         select *
         from {{ ref('ledger_fee_stats_agg') }}
-        where
-            day_agg < date('{{ var("batch_end_date") }}')
-            {% if is_incremental() %}
-                and day_agg >= date('{{ var("batch_start_date") }}')
-            {% endif %}
     )
 
     , final as (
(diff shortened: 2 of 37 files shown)


models/marts/daily_fee_stats_agg.sql
```diff
@@ -21,11 +20,6 @@ with
     ledger_stats as (
         select *
         from {{ ref('ledger_fee_stats_agg') }}
```
hunterpack, line 22: removing the `day_agg < date(batch_end_date)` filter changes what ends up in this table. That filter ran on every build, not just incremental ones, and it kept out the current day. Without it, the table will now include a row for today that keeps growing as the day goes on. I checked prod: `ledger_fee_stats_agg` already has data for today, while this table stops at yesterday.

Just want to confirm this is intended

models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
```diff
@@ -1,17 +1,17 @@
-{% set batch_size = 'year' if flags.FULL_REFRESH else 'day' %}
+{% set batch_size = env_var('DBT_MICROBATCH_BATCH_SIZE', '') or ('year' if flags.FULL_REFRESH else 'day') %}
```
hunterpack: Nit: Can you add a note about `DBT_MICROBATCH_BATCH_SIZE` somewhere? It's a new env var that changes how backfills run, but it's not in the PR description or any docs, so might be good to call out or document somewhere. 

Also, the same line is copy-pasted into 14 files here and 8 more in stellar-dbt. Might be worth a small macro in this repo (similar to how stellar-dbt handles microbatch_begin()) so it lives in one place and both repos can use it.

hunterpack (approved):
The description lists trade_agg under "to table", but trade_agg.sql on this branch is still incremental merge with unique_key and appears in neither PR's diff. Missing change or stale description?

Other than that and these few comments looking good

models/marts/daily_fee_stats_agg.sql
amishas157: Yes, this is expected. For WAP partial refresh/ WAP incremental run, we will actually build data in different dataset say prod_backfill and then copy over partitions. 

For table materialization, we have to copy entire table as the whole table is built again.
For incremental materialization, since only few partitions are built again and only those will be copied over.

If we keep batch_end_date with table , there is a risk when partial refreshing datasets. partial refresh will refresh table only until batch_end_date and then swap it. Future data will be blown up by this.

Yes, there is a trade off that these table will start having data for today, which could also be partial. But in the next run it eventually fixes.

amishas157:
> The description lists trade_agg under "to table", but trade_agg.sql on this branch is still incremental merge with unique_key and appears in neither PR's diff. Missing change or stale description?
> 
> Other than that and these few comments looking good

Yes, kept trade_agg as is because there were [other issues](https://github.com/stellar/stellar-dbt-public/issues/305) with it. Updated description.

models/intermediate/account_balances/int_account_balances__liquidity_pools.sql
amishas157: Good idea. Implemented in https://github.com/stellar/stellar-dbt-public/pull/302/changes/97915ed61c6a353ee4deedb4dc4126c48951fa17

PR #310, opened 2026-07-31
snapshots: compute a day range in one set-based pass

amishas157:
## What

`calculate_snapshot_diff_for_day_range` looped over every day in the requested range, emitting three statements per day. This replaces the loop with a single set-based computation. Statement count is now fixed at two regardless of whether the range is one day or several years.

The macro signature is unchanged, so `incremental_snapshot.sql` is untouched and **no snapshot model changes are needed** — the eight snapshots here and the three in `stellar-dbt` are unaffected.

## Why it works

SCD Type-2 `valid_to` is always the `valid_from` of the entity's next version. `calculate_snapshot_diff_for_day` never closes a row without a successor (set 3 requires a matching new source row), and it emits one version per entity per day with the last source row of that day winning. So `valid_to` is exactly `LEAD(valid_from)` over that entity's ordered versions — no day in the range depends on the previous day's output, and the whole range chains with one window function.

## The correctness fix

The day loop seeded its first iteration from the target's rows `WHERE valid_to IS NULL`. That is the state as of *now*, not as of `snapshot_start_date`, which is why re-running a historical range could not reproduce it (stellar/stellar-dbt#733).

Each entity's version already open at the range start is now read from the target **positionally** — its latest version with `valid_from` before `snapshot_start_date`. The additional `valid_to IS NULL OR valid_to >= start` predicate on that read is redundant for correctness and present only to prune the `valid_to`-partitioned target.

## Two statements

1. `temp_source_table` — one row per entity per day in the range, with `valid_from` set.
2. `temp_target_table` — those rows unioned with the versions open at the range start, `valid_to` chained across both sets at once. Open-version rows that receive no `valid_to` belong to entities with no activity in the range and are dropped, so untouched entities contribute nothing to the diff.

## Notes for review

- **Open versions are chained via `PARTITION BY`, not a join.** The unique key may contain nulls — a trustline has either `liquidity_pool_id` or `asset_code`/`asset_issuer` set, never both, and `stg_trust_lines` does not coalesce them. `PARTITION BY` treats nulls as equal; `=` does not. Worth checking separately whether the existing equality joins in `calculate_snapshot_diff_for_day` and `merge_source_into_target` have been duplicating versions for those keys — that is a live-data question, not answerable from the code.
- **Target columns are aligned to the source column list**, with `CAST(NULL AS <type>)` for columns the target lacks. The per-day merge absorbed schema drift via `dest_cols_csv`; a `UNION ALL` does not. This matters because `process_schema_changes` adds new columns to the target only *after* the presql runs.
- **Open-version rows carry their full payload**, because the materialization's merge on `(source_unique_key, valid_from)` updates all columns on match.
- Added a compiler error when `snapshot_start_date >= snapshot_end_date`. Previously that produced an empty range silently.
- `calculate_snapshot_diff_for_day` is deliberately left in place, now unreferenced, to serve as the reference implementation while equivalence fixtures are written. It is removed in a follow-up.

## Validation

`macros/` is in `.sqlfluffignore`, and both sqlfluff pre-commit hooks fail identically on untouched files in this repo (macOS `sed` in `pre-commit/for_pre_commit.sh`, and `DBT_TARGET` not reaching dbt at hook time). All other hooks pass. Not yet run against BigQuery.

Validated offline instead:

- **Rendered the macro** with stubbed dbt globals against a `stg_trust_lines`-shaped schema for the incremental, full-refresh, and inverted-range cases. This caught a real defect: a `{#-- --#}` comment stripped whitespace on both sides and glued `AS target_table` to `WHERE`. Fixed.
- **Ran the emitted query's logic** on a four-entity scenario covering: an entity with a missing version inside the range (recovered, and its predecessor closed against it), a brand-new entity (chain built from scratch), an entity with no activity in the range (entirely absent from the diff), and null unique-key components throughout. `valid_to = LEAD(valid_from)` holds across the result.

## Not addressed here

- The diff still only inserts and updates, so a target row whose source row has since **disappeared** is not removed. Unchanged from current behaviour, so no regression, but it means a re-run repairs missing and changed versions and not extra ones. Documented in the macro and in `docs/snapshot.md` limitations. Needs the rollback step — stellar/stellar-dbt#988.
- Chunked batching and key-scoped (partial) refreshes.
- The equivalence fixtures against the `dummy_snapshot_*` harness — both engines over one `(source_data, target_data, expected_target)` triple, plus genesis-recompute equals `--full-refresh`. These should land before this is trusted on a large snapshot.
- With the day loop gone, the presql hook that justified a custom materialization is gone too. `incremental_snapshot.sql` is a fork of dbt-bigquery's incremental materialization that has drifted from it — no grants, no view→table guard, no `microbatch`, no `copy_partitions`, no `_dbt_max_partition`. Un-forking it is worth its own change.

Related: stellar/stellar-dbt#733, stellar/stellar-dbt#988


amishas157:
Pushed `9328dea`. Writing tests against the existing fixture harness found two real defects in the first commit, so flagging what changed.

### Two defects the existing `dummy_snapshot_incremental` fixture caught

1. **A version starting exactly on `snapshot_start_date` was never re-closed.** The boundary predicate was `valid_from < start`; fixture account 2 has `valid_from = 2025-01-01 00:00:00` with the range starting `2025-01-01`, so its open version stayed open while a later version was inserted in front of it. Two open versions for one entity.
2. **A target version newer than a source row in the range was not chained.** Fixture account 3. The source version got inserted unlinked, again leaving two open versions.

Fixed by reading the target's **in-range** versions as well, and chaining all three sets — source versions, the pre-range version, in-range target versions — through one `LEAD`. A source version supersedes the target copy at the same `valid_from`, so no duplicate `valid_from` reaches the merge. Entities with no source activity in the range are still excluded, so they stay untouched.

### One deliberate behaviour change, needs a call

Defect 2's old behaviour was to **drop** the source version. That is recorded as intended — `docs/snapshot.md` lists it under limitations ("slight chance of missing history ... when a new version of record is already present") and the fixture's own description said "record to be dropped as new version exist in target".

The range is now rebuilt from every version that belongs in it, so that earlier version is recovered instead of discarded. I think that is right — it is the same missing-history problem stellar/stellar-dbt#733 is about, and silently dropping source history is what makes re-runs unreproducible. But it changes a documented expectation, so it should be an explicit decision rather than something noticed in review. `dummy_snapshot_incremental`'s expected data gains that row.

### New fixtures

- `dummy_snapshot_rerun` — re-running a range the snapshot already covers must leave the target byte-identical. This is the property the day-by-day implementation could not provide, and the one #733 asks for.
- `dummy_snapshot_null_unique_key` — composite `source_unique_key` with nulls in a key column, as `trustlines_snapshot` has. Matching such keys with `=` fails on the null column and inserts a second open version rather than closing the existing one.

### Wiring needed in stellar-dbt

`stellar-dbt`'s `ci.yml` invokes `run_snapshot_test.sh` once per fixture by name, so the two new fixtures need a step each there, plus adding to the `dummy_snapshot` exclusion lists in `lint-and-test.yml`, `upload-manifest.yml`, `upload-release-manifest.yml` and the other `ci.yml` selectors. Happy to open that PR.

### Verification

Still not run against BigQuery. Offline, mirroring the emitted CTE structure: both existing fixtures pass, the two new ones pass, plus a same-range re-run, a narrower sub-range re-run after the head moved past it, and the null-key trustlines example — 7 cases, each also asserting `valid_to = lead(valid_from)` across the whole target. `run_snapshot_test.sh dummy_snapshot_incremental` from `stellar-dbt` on the `ci` target is the real check.

amishas157:
Pushed `0495839` — adds `snapshot_rollback`, which closes the two cases the additive diff could not.

## What it does

Off by default. When set, the source becomes authoritative for `[snapshot_start_date, snapshot_end_date)`: target versions whose `valid_from` falls inside the range are left out of the rebuilt chain and deleted. That resolves both remaining cases — a source row that has disappeared, and a source row that moved to a different timestamp on the same day (which is the same defect, since both leave a version the source no longer produces).

Versions at or after `snapshot_end_date` are outside the recomputed range, so they are preserved and stitched onto the last rebuilt version. A bounded range does not disturb later history, which means chunked backfills are safe.

Two details worth review:

- **Entities with no source activity in the range are normally excluded from the diff entirely.** In rollback mode, one whose in-range versions are being deleted still has to be rebuilt — otherwise its version *before* the range keeps a `valid_to` pointing at a deleted row. That is what `in_range_target_versions_for_entity` is for, and `dummy_snapshot_rollback` account 2 covers it.
- **Statement ordering.** The delete runs after the diff is built (the diff reads those rows to decide which entities need rebuilding) and before the merge (so rows the merge inserts are not deleted again). Getting this backwards silently loses the new versions.

It reads from model config before the var, so fixtures can set it per model while a backfill passes `--vars '{"snapshot_rollback": true}'` with no model changes. It raises a compiler error against the `prod` target since it issues DELETEs.

## Removed the null-key fixture

`dummy_snapshot_null_unique_key` from the previous commit is gone. It cannot pass: `default__get_merge_sql` builds its `ON` predicate with plain `=` for **list** `unique_key`s, with no dispatch to `bigquery__get_merge_unique_key_match`, so a NULL key component never matches and the row is inserted rather than updated.

That is a pre-existing problem independent of this PR, and `trustlines_snapshot` is the one snapshot exposed to it. Filed as #311 with a query to confirm whether it is already happening in prod. Leaving it out keeps this PR green rather than red for a reason that has nothing to do with the rewrite.

## Verification

Offline, mirroring the emitted CTE structure — 9 cases now:

- both existing fixtures, `dummy_snapshot_rerun`, and the trustlines worked example
- same-range re-run and narrower sub-range re-run, both byte-identical
- rollback: stale version removed for an entity with source activity; stale version removed and predecessor reopened for an entity with *no* source activity; version at the old timestamp removed after a same-day shift
- rollback re-run byte-identical
- after-range version preserved and stitched under a bounded rollback
- rollback OFF still produces the committed fixture expectations, unchanged

Each also asserts `valid_to = lead(valid_from)` across the whole target. The macro renders cleanly in both modes and the prod guard fires.

Still not run against BigQuery. `run_snapshot_test.sh dummy_snapshot_rollback` from `stellar-dbt` on the `ci` target is the real check, and it needs a step in that repo's `ci.yml` — same wiring as `dummy_snapshot_rerun`.

PR #312, opened 2026-08-04
Refactor snapshot macro to support backfills

amishas157:
This PR refactors snapshot macro to:
- Make snapshots building faster
- Support backfills with WAP
- Make snapshots idempotent

**Before this PR:**

Snapshots for a day range was calculated by:

```
-- DAY 1
CREATE ....;
UPDATE ....;
MERGE INTO ... target_temp_table;

-- DAY 2
CREATE ....;
UPDATE ....;
MERGE INTO ... target_temp_table;
```

The ☝️ create 3 DML statement for each day and was not super efficient.

Backfill / Backfixing required manual deletion of ingested data and therefore was unintuitive.

**After this PR**

Snapshots for a day range is calculated by:

```
-- step 1, once: return the target to its state at snapshot_start_date
DELETE ... WHERE valid_from >= snapshot_start_date;
UPDATE ... SET valid_to = NULL WHERE valid_to >= snapshot_start_date;

-- days between snapshot_start_date and snapshot_end_date are batched in 3 months chunks
-- if snapshot_end_date if not present, chunks are created until current_date

-- step 2, per chunk of snapshot_start_date/snapshot_end_date (3 months).
CREATE ... source_temp_table;   -- the chunk's versions (chained using lead function)
CREATE ... target_temp_table;   -- those versions chained onto the one current at the chunk start
MERGE  ... INTO target;         -- applied before the next chunk begins

Contd. until all chunks are finished
```

The ☝️ creates 3 DML statement per chunk i.e. per 3 months and therefore runs very fast as compared to previous approach.

**What remains same**

Repairing a hole is not possible as it would require fixing boundaries on two side and can pose some risks. However, when a user kicks off a full refresh/ partial refresh, materialization auto takes care of deletion and recreation of data.
Repairing means that snapshot should always refresh until today

**Backfill consistency with microbatch models**

Microbatch models have a `begin` date which indicates when the full refresh should happen from. Similarly, I baked a default start date in snapshot models to replicate the behaviour in case of full refresh.
Snapshots models are also backfilled in batches, however the chunk size(3 months) is constant in snapshot tables and is processed serially unlike microbatches.

**Support to backfill only certain assets**

We always had a requirement to partially refresh snapshots, for example when a new recognized asset is added and it's pricing is backfilled as well.

This PR adds support to backfill snapshot tables using snapshot keys

```
dbt run --select contract_data_snapshot \
  --vars '{"snapshot_start_date": "2024-01-01", "snapshot_keys": ["CONTRACT_A", "CONTRACT_B"]}'
```

**Airflow changes**
https://github.com/stellar/stellar-etl-airflow/pull/985 updates backfill dag to:
- For full refresh do not send start date/end date  - this use snapshot models default start date and current date as end date
- For partial refresh, send start date - this use given start date and current date as end date
- Adds support to send snapshot keys during snapshot backfills

**Testing**

[Kicked off full backfill](https://4c03f3f96a2347c6bdb4cff6041c4a23-dot-us-central1.composer.googleusercontent.com/dags/dbt_backfill/grid?dag_run_id=manual__2026-08-07T03%3A45%3A27%2B00%3A00&tab=details) for all 11 snapshot models in staging DAG. They finished in **1 hour 15 mins** with no requirement of upping the slots or using autoscaled slots. 

Datadiffs were also verified

**Unrelated changes**

This PR gets rid off dummy snapshots which were initially added as unit tests for snapshots macros, however they add more complexity than ensuring regression tests.

**Tech debt which still remains**

Get rid off snapshot start/end date and snapshot full refresh variables and use existing variables `batch_start_date, batch_end_date, full_refresh`. Leaving it out of scope of this work


amishas157:
Rendered SQL for a two-chunk incremental run, so this can be reviewed as SQL rather than as Jinja.

`evicted_keys_snapshot`, `snapshot_start_date=2025-09-03`, `snapshot_end_date=2026-01-05`, default 3-month chunks.

```
Rebuilding evicted_keys_snapshot from 2025-09-03 to 2026-01-05 in 2 chunk(s) of 3 month(s)
```

Chunk windows: `[2025-09-03, 2025-12-01)`, `[2025-12-01, 2026-01-05)`

---

### Statement 1 — reset (`snapshot_reset_from_start`), once

```sql
delete from `sh`.`staging_snapshots`.`evicted_keys_snapshot`
    where valid_from >= timestamp(date('2025-09-03'))
        and (
            valid_to is null
            or valid_to >= timestamp(date('2025-09-03'))
        )
        ;

    update `sh`.`staging_snapshots`.`evicted_keys_snapshot`
    set valid_to = null
    where valid_to >= timestamp(date('2025-09-03'))
        ;
```

---

### Chunk 1/2 — `[2025-09-03, 2025-12-01)`

```sql

CREATE OR REPLACE TABLE `sh`.`staging_snapshots`.`evicted_keys_snapshot_source`
OPTIONS(expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour))
AS (
    with source_ranked as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , row_number() over (
                partition by `ledger_key_hash`, date(closed_at)
                order by closed_at desc
            ) as row_num
        from `sh`.`staging`.`stg_restored_key`
        where date(closed_at) >= date('2025-09-03')
            and date(closed_at) < date('2025-12-01')
            
    )

    select
        `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
        , cast(closed_at as timestamp) as valid_from
    from source_ranked
    where row_num = 1);



CREATE OR REPLACE TABLE `sh`.`staging_snapshots`.`evicted_keys_snapshot_target` PARTITION BY valid_to
OPTIONS(expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour))
AS (
    with source_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , 'source' as row_origin
        from `sh`.`staging_snapshots`.`evicted_keys_snapshot_source`
    )

    , boundary_candidates as (
        select
             target_table.`ledger_key_hash`
            , target_table.`is_evicted`
            , target_table.`ledger_sequence`
            , target_table.`closed_at`
            , target_table.valid_from
            , row_number() over (
                partition by
                     target_table.`ledger_key_hash`
                order by target_table.valid_from desc
            ) as row_num
        from `sh`.`staging_snapshots`.`evicted_keys_snapshot` as target_table
        
        where target_table.valid_from < timestamp(date('2025-09-03'))
            and (
                target_table.valid_to is null
                or target_table.valid_to >= timestamp(date('2025-09-03'))
            )
            
    )

    , boundary_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , 'boundary' as row_origin
        from boundary_candidates
        where row_num = 1
    )

    , all_versions as (
        select * from source_versions
        union all
        select * from boundary_versions
    )

    , flagged_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , countif(row_origin = 'source') over (
                partition by `ledger_key_hash`
            ) as source_versions_for_entity
        from all_versions
    )

    select
        `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
        , valid_from
        , lead(valid_from) over (
            partition by `ledger_key_hash`
            order by valid_from
        ) as valid_to
    from flagged_versions where source_versions_for_entity > 0);
```

Then dbt's own merge on `(ledger_key_hash, valid_from)`, from `bq_generate_incremental_build_sql` — unchanged from before this PR.

---

### Chunk 2/2 — `[2025-12-01, 2026-01-05)`

```sql

CREATE OR REPLACE TABLE `sh`.`staging_snapshots`.`evicted_keys_snapshot_source`
OPTIONS(expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour))
AS (
    with source_ranked as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , row_number() over (
                partition by `ledger_key_hash`, date(closed_at)
                order by closed_at desc
            ) as row_num
        from `sh`.`staging`.`stg_restored_key`
        where date(closed_at) >= date('2025-12-01')
            and date(closed_at) < date('2026-01-05')
            
    )

    select
        `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
        , cast(closed_at as timestamp) as valid_from
    from source_ranked
    where row_num = 1);



CREATE OR REPLACE TABLE `sh`.`staging_snapshots`.`evicted_keys_snapshot_target` PARTITION BY valid_to
OPTIONS(expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour))
AS (
    with source_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , 'source' as row_origin
        from `sh`.`staging_snapshots`.`evicted_keys_snapshot_source`
    )

    , boundary_candidates as (
        select
             target_table.`ledger_key_hash`
            , target_table.`is_evicted`
            , target_table.`ledger_sequence`
            , target_table.`closed_at`
            , target_table.valid_from
            , row_number() over (
                partition by
                     target_table.`ledger_key_hash`
                order by target_table.valid_from desc
            ) as row_num
        from `sh`.`staging_snapshots`.`evicted_keys_snapshot` as target_table
        
        where target_table.valid_from < timestamp(date('2025-12-01'))
            and (
                target_table.valid_to is null
                or target_table.valid_to >= timestamp(date('2025-12-01'))
            )
            
    )

    , boundary_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , 'boundary' as row_origin
        from boundary_candidates
        where row_num = 1
    )

    , all_versions as (
        select * from source_versions
        union all
        select * from boundary_versions
    )

    , flagged_versions as (
        select
            `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
            , valid_from
            , countif(row_origin = 'source') over (
                partition by `ledger_key_hash`
            ) as source_versions_for_entity
        from all_versions
    )

    select
        `ledger_key_hash`, `is_evicted`, `ledger_sequence`, `closed_at`
        , valid_from
        , lead(valid_from) over (
            partition by `ledger_key_hash`
            order by valid_from
        ) as valid_to
    from flagged_versions where source_versions_for_entity > 0);
```

Then dbt's own merge on `(ledger_key_hash, valid_from)`, from `bq_generate_incremental_build_sql` — unchanged from before this PR.

amishas157:
## Key-filtered run against staging

Same pipeline with `snapshot_keys` set — from an actual run, not a rendering.
`wisdom_tree_asset_prices_data_v2` on staging-hubble, scoped to one asset.

### dbt output

```
Rebuilding wisdom_tree_asset_prices_data_v2 from 2026-01-01 to 2026-08-05 in 3 chunk(s) of 3 month(s) for selected keys only
Reset wisdom_tree_asset_prices_data_v2 to its state at 2026-01-01
Chunk 1/3 done: wisdom_tree_asset_prices_data_v2 rebuilt through 2026-04-01 (chunk covered 2026-01-01 to 2026-04-01)
Chunk 2/3 done: wisdom_tree_asset_prices_data_v2 rebuilt through 2026-07-01 (chunk covered 2026-04-01 to 2026-07-01)
Chunk 3/3 done: wisdom_tree_asset_prices_data_v2 rebuilt through 2026-08-05 (chunk covered 2026-07-01 to 2026-08-05)
```

### Statements executed, in order

| # | statement | chars | filter predicates |
|---|---|---|---|
| 1 | reset — delete + reopen | 603 | **2** |
| 2 | chunk 1 — diff (both temp tables) | 4874 | **2** |
| 3 | chunk 1 — tmp relation + merge | 1764 | **0** |
| 4 | chunk 2 — diff (both temp tables) | 4874 | **2** |
| 5 | chunk 2 — tmp relation + merge | 1764 | **0** |
| 6 | chunk 3 — diff (both temp tables) | 4874 | **2** |
| 7 | chunk 3 — tmp relation + merge | 1764 | **0** |

Seven statements: one reset, then two per chunk. The filter appears **twice in the reset** (delete and reopen) and **twice in every chunk diff** (source read and boundary read) — the four sites it must reach. The merges carry none, which is correct: they operate on an already-scoped diff table and there is no `not matched by source` clause, so nothing unfiltered can be touched.

---

### Statement 1 — reset

The filter is on **both** statements. Missing from the second, unrelated assets would be spuriously reopened.

```sql
delete from `staging-hubble`.`staging_snapshots`.`wisdom_tree_asset_prices_data_v2`
    where valid_from >= timestamp(date('2026-01-01'))
        and (
            valid_to is null
            or valid_to >= timestamp(date('2026-01-01'))
        )
         and `asset_contract_id` in ('CDQULNXLNTST3ERDKI2UXFUL64WGVJA4INLSYZCQLBI5KHQ4PSLFJI5S');
    update `staging-hubble`.`staging_snapshots`.`wisdom_tree_asset_prices_data_v2`
    set valid_to = null
    where valid_to >= timestamp(date('2026-01-01'))
         and `asset_contract_id` in ('CDQULNXLNTST3ERDKI2UXFUL64WGVJA4INLSYZCQLBI5KHQ4PSLFJI5S');
```

---

### Statement 2 — chunk 1 diff, `[2026-01-01, 2026-04-01)`

Filter on the source read, and again on the boundary read via the `target_table` alias.

```sql
CREATE OR REPLACE TABLE staging-hubble.staging_snapshots.wisdom_tree_asset_prices_data_v2_source
    OPTIONS(
        expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour)
    )
    AS (
    with source_ranked as (
        select
            `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
            , row_number() over (
                partition by `asset_contract_id`, date(day)
                order by day desc
            ) as row_num
        from `staging-hubble`.`staging_sdf_raw`.`stg_wisdom_tree_asset_prices_data`
        where date(day) >= date('2026-01-01')
            and date(day) < date('2026-04-01')
             and `asset_contract_id` in ('CDQULNXLNTST3ERDKI2UXFUL64WGVJA4INLSYZCQLBI5KHQ4PSLFJI5S')
    )
    select
        `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
        , cast(day as timestamp) as valid_from
    from source_ranked
    where row_num = 1);
CREATE OR REPLACE TABLE staging-hubble.staging_snapshots.wisdom_tree_asset_prices_data_v2_target
    partition by timestamp_trunc(valid_to, month)OPTIONS(
        expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 3 hour)
    )
    AS (
    with source_versions as (
        select
            `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
            , valid_from
            , 'source' as row_origin
        from staging-hubble.staging_snapshots.wisdom_tree_asset_prices_data_v2_source
    )
    , boundary_candidates as (
        select
             target_table.`batch_id`
            , target_table.`batch_run_date`
            , target_table.`batch_insert_ts`
            , target_table.`asset_code`
            , target_table.`asset_issuer`
            , target_table.`asset_type`
            , target_table.`close_usd`
            , target_table.`day`
            , target_table.`airflow_start_ts`
            , target_table.`asset_contract_id`
            , target_table.valid_from
            , row_number() over (
                partition by
                     target_table.`asset_contract_id`
                order by target_table.valid_from desc
            ) as row_num
        from `staging-hubble`.`staging_snapshots`.`wisdom_tree_asset_prices_data_v2` as target_table
        where target_table.valid_from < timestamp(date('2026-01-01'))
            and (
                target_table.valid_to is null
                or target_table.valid_to >= timestamp(date('2026-01-01'))
            )
             and target_table.`asset_contract_id` in ('CDQULNXLNTST3ERDKI2UXFUL64WGVJA4INLSYZCQLBI5KHQ4PSLFJI5S')
    )
    , boundary_versions as (
        select
            `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
            , valid_from
            , 'boundary' as row_origin
        from boundary_candidates
        where row_num = 1
    )
    , all_versions as (
        select * from source_versions
        union all
        select * from boundary_versions
    )
    , flagged_versions as (
        select
            `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
            , valid_from
            , countif(row_origin = 'source') over (
                partition by `asset_contract_id`
            ) as source_versions_for_entity
        from all_versions
    )
    select
        `batch_id`, `batch_run_date`, `batch_insert_ts`, `asset_code`, `asset_issuer`, `asset_type`, `close_usd`, `day`, `airflow_start_ts`, `asset_contract_id`
        , valid_from
        , lead(valid_from) over (
            partition by `asset_contract_id`
            order by valid_from
        ) as valid_to
    from flagged_versions where source_versions_for_entity > 0);
    create or replace table `staging-hubble`.`staging_snapshots`.`wisdom_tree_asset_prices_data_v2__dbt_tmp172248767312`
    partition by timestamp_trunc(valid_to, month)
    OPTIONS(
      description="""This table captures historical snapshots of price of wisdom tree assets in USD(Sourced from RWA). The valid_from and valid_to timestamps help in identifying when each price record was active.""",
      expiration_timestamp=TIMESTAMP_ADD(CURRENT_TIMESTAMP(), INTERVAL 12 hour)
    )
    as (
      -- depends_on: `staging-hubble`.`staging_sdf_raw`.`stg_wisdom_tree_asset_prices_data`
SELECT * from staging-hubble.staging_snapshots.wisdom_tree_asset_prices_data_v2_target
    );
```

---

Chunks 2 and 3 are identical but for their window. To pull the full text from a run:

```bash
python3 -c "
import re
c = re.sub(r'\x1b\[[0-9;]*m','',open('logs/dbt.log',errors='replace').read())
n = 'model.stellar_dbt.wisdom_tree_asset_prices_data_v2'
p = r'On '+re.escape(n)+r': (?:/\*.*?\*/)?\s*\n(.*?)(?=\n\d\d:\d\d:\d\d\.\d+ \[)'
for m in re.finditer(p, c, re.S):
    print('='*70); print(m.group(1).strip())
"
```

macros/materializations/incremental_snapshot.sql
```diff
@@ -1,130 +1,247 @@
...
+--
+-- To rebuild only part of the history on purpose, use an ordinary run with an early start
+-- rather than a full refresh -- the reset deletes from that date forward regardless.
+--
+-- end: defaults to now, and callers should leave it that way. snapshot_end_date overrides it,
+-- but the reset below is not bounded by it: step 1 deletes every version from start_date
+-- onward while the chunks only rebuild as far as the end date, so anything between that date
+-- and now is deleted and never recreated.
+--#}
+{%- set requested_start_date = config.get('snapshot_start_date') or var('snapshot_start_date', none) -%}
```
chowbao: Can you do the start date trunc here instead of https://github.com/stellar/stellar-etl-airflow/pull/985/changes#diff-eaf3341bcf1a02e40e75b38f1232e4504796e7ea0e4bd16bd387be75d531a15aR175

```
"ti.xcom_pull(task_ids='{resolve_id}')['event_time_start'][:10]"
```

The airflow var setting seems more like a hack instead of just being able to pass the general timestamp

macros/materializations/incremental_snapshot.sql
```diff
@@ -1,130 +1,247 @@
...
+        ~ "snapshot_start_date is defined as a project var, so a var(..., 'default') fallback in "
+        ~ "the model is never reached -- put the model's own date in snapshot_default_start_date "
+        ~ "instead, which a full refresh will use."
+    ) -%}
+  {%- endif -%}
+{%- endif -%}
+
+{%- set end_date = config.get('snapshot_end_date')
+    or var('snapshot_end_date', none)
+    or run_started_at.strftime('%Y-%m-%d') -%}
```
chowbao: Wouldn't you want the var `snapshot_end_date` to have the highest priority because that is what airflow would pass in?

I see that the config `snapshot_end_date` just reads the var `snapshot_end_date` anyways which kinda makes this end date logic a little circular/redundant

macros/materializations/incremental_snapshot.sql
```diff
@@ -1,130 +1,247 @@
...
+--
+-- end: defaults to now, and callers should leave it that way. snapshot_end_date overrides it,
+-- but the reset below is not bounded by it: step 1 deletes every version from start_date
+-- onward while the chunks only rebuild as far as the end date, so anything between that date
+-- and now is deleted and never recreated.
+--#}
+{%- set requested_start_date = config.get('snapshot_start_date') or var('snapshot_start_date', none) -%}
+{%- set default_start_date = config.get('snapshot_default_start_date', none) -%}
+
+{%- if full_refresh_mode -%}
```
chowbao, line 63: I think we should include a check to not have `snapshot_keys` set when a full-refresh is being run. I'm not 100% sure but I think this would then refresh the whole table with only the snapshot_keys instead of the whole table cause a full refresh would delete all rows of the table rather than only the matching `snapshot_key` rows

models/snapshots/accounts_snapshot.sql
```diff
@@ -1,8 +1,15 @@
 -- depends_on: {{ ref('stg_accounts') }}
 {%- set temp_source_table = this.table ~ '_source' -%}
 {%- set temp_target_table = this.table ~ '_target' -%}
+{%- set snapshot_default_start_date = snapshot_begin('2015-09-30') -%}
 
 {% set meta_config = {
+    "datadiff": {
+        "unique_key": ["account_id", "valid_from"],
+        "exclude_columns": ["batch_id", "batch_run_date", "batch_insert_ts", "airflow_start_ts"],
+        "min_match_percent": 98,
```
chowbao, line 10: I wonder if it makes sense to make this a `var(match_pct, 98)`

I feel like there's gonna be times where we have exceptions to the 98% threshold and I wonder if it makes sense to just make this an adjustable var rather than needing to edit dbt sql

chowbao:
**nit (non-blocking):** a few cleanup opportunities while we're in here — mostly dead code left behind by the refactor and repeated boilerplate that could live in the materialization instead of every model.

If not fixing now, let's create a github issue to track the tech debt (plus `Get rid off snapshot start/end date and snapshot full refresh variables and use existing variables batch_start_date, batch_end_date, full_refresh. Leaving it out of scope of this work`) from the PR description

**Dead code / stale docs**
- `macros/generate_date_range.sql` — zero callers now that `calculate_snapshot_diff_for_day_range` is deleted; can be removed.
- `macros/incremental_snapshot_utils.sql` — `merge_source_into_target` and `get_active_records_from_target_table` also have no callers after this PR; only `create_temp_table_with_data` is still used.
- `.claude/CLAUDE.md` L71-72 still documents `calculate_snapshot_diff_for_day` / `_for_day_range`, which this PR deletes.
- `docs/snapshot.md` L203 tells users to "call `create_snapshot` macro directly" — that macro doesn't exist in the repo.

**Model config boilerplate** — each snapshot model repeats lines the materialization could own, so every new snapshot model has to copy them correctly:
- `"snapshot_start_date": var(...)` / `"snapshot_end_date": var(...)` are dead — the materialization already falls back to the same vars (`config.get(...) or var(..., none)`), and its own error message notes the model-level fallback can never fire.
- `temp_source_table` / `temp_target_table` are `this.table ~ '_source'/'_target'` in every model — could default in the materialization: `config.get('temp_source_table') or this.table ~ '_source'`.
- `"full_refresh": true if var(...) == 'true' else none` — could move into the materialization (`full_refresh_mode = should_full_refresh() or var('snapshot_full_refresh', 'false') == 'true'`) so one place decides. Also worth noting the switch from a computed `false` to `none` re-enables the CLI `--full-refresh` flag for these models (dbt only defers to the flag when config is `none`); if that's intentional it'd be good to say so, since blanket `--full-refresh` runs now trigger a genesis rebuild.

Together those drop each model's config from ~15 boilerplate lines to ~8 model-specific ones.

**Smaller ones**
- `snapshot_begin` is byte-identical to the existing `microbatch_begin` except the var name and a required-default guard — models could call `microbatch_begin('2015-09-30')`, or it could grow a `var_name` param. `snapshot_ci_window_days` has no setters yet.
- The five new macro `.yml` descriptions restate the `.sql` header comments nearly verbatim — two copies of the same prose to keep in sync; one-liners would do with the header as source of truth.
- `incremental_snapshot.sql` L111-115: the "built once… callers must not re-derive it" comment is contradicted by `calculate_snapshot_diff`, which re-derives the filter itself for the aliased boundary read — either drop the comment or thread an aliased variant.
- `snapshot_chunk_ranges`: the `max_chunks` upper bound + `cursor < end` guard is a while-loop workaround; computing the chunk count and looping `range(n)` drops the namespace and guard.
- `snapshot_chunk_months` is settable via both config and var; nobody sets either — the var alone seems enough.

macros/materializations/incremental_snapshot.sql
amishas157: The check exists in airflow pr https://github.com/stellar/stellar-etl-airflow/blob/snapshot-start-one-day-earlier/dags/stellar_etl_airflow/build_resolve_models_task.py#L239-L246

It is assumed that we are only going to backfill using backfill dag and the dag planner will catch it at airflow level

models/snapshots/accounts_snapshot.sql
amishas157: I believe it it is difficult to calculate a right matching percentage threshold for data-diffs. The remaining 2 percent is to account for minor changes i.e decimal tolerance etc. The current expectation is that backfill operator should observe the diffs manually only except for the cases of precision diffs, etc.
We can discuss how more in details in ofc hours how to make this part more independent and then I can wrap configurable matching percentage as separate PR

macros/materializations/incremental_snapshot.sql
amishas157: Agree, there is opportunity of cleanup here.

amishas157: Addressed in https://github.com/stellar/stellar-dbt-public/pull/312/commits/685d1fc99d337b289a9d92ff907c65dc4c24f2bf

macros/materializations/incremental_snapshot.sql
amishas157: Addressed in https://github.com/stellar/stellar-dbt-public/pull/312/commits/685d1fc99d337b289a9d92ff907c65dc4c24f2bf

amishas157:
Addressed in 685d1fc9:

- **r3751796535** — the materialization truncates both dates now, so a caller can pass a full timestamp; stellar-etl-airflow can drop its `[:10]` once this is pinned.
- **r3751934004** — end date reads config only: `config.get("snapshot_end_date") or today`. Same for the start.
- Dead macros removed: `generate_date_range`, `merge_source_into_target`, `get_active_records_from_target_table`.
- Stale docs fixed: the `CLAUDE.md` macro list, the `create_snapshot` reference, and `snapshot_key_filter.yml`, which still said composite keys were unsupported.
- Macro `.yml` descriptions no longer restate their `.sql` headers.

Also clarified that `snapshot_end_date` bounds the rebuild but not the delete, so a past end date truncates rather than repairs — hole repair stays unsupported.

Keeping the per-model config boilerplate deliberately.

macros/materializations/incremental_snapshot.sql
chowbao: That's fine. We should just let the team know about this weird edge case in case they have claude running just dbt locally and not just using airflow UI directly

chowbao (approved):

chowbao:
FYI this currently targets master instead of a release branch

PR #313, opened 2026-08-17
Add a single exceptions table for known test failures

amishas157:
Same idea as stellar-dbt#1102, but for this public package's own tests. Also moves two exception rules that were hardcoded inside test SQL files into this same table, so they're visible instead of buried in code.


amishas157:
Pushed two more commits consolidating this mechanism so stellar-dbt can call it directly instead of forking it (per discussion in stellar/stellar-dbt#1102):

**`ae02e43b`** — `exclude_test_exceptions()` and a new `validate_test_exceptions()` (extracted from `tests/test_exceptions_valid.sql`) now take the registry and the exceptions-seed relation as **explicit arguments** instead of looking either up by name internally. This isn't just tidier — it's necessary: dbt's macro-override rule means an unqualified macro call resolves against the ROOT project's definition of that name, even when the call happens inside this package's own code. If these macros had kept calling a no-arg `test_exception_targets()` / hardcoded `ref('public_test_exceptions')` internally, a downstream project (stellar-dbt) defining its own same-named registry macro would silently override the lookup used by **this package's own tests** whenever they run as a dependency inside that project's build — exactly the wrong-registry bug this mechanism exists to prevent. `test_exception_targets()` itself stays local and unshared, since each project's set of wired tests is genuinely different data.

**`e00e4903`** — a real bug caught while verifying the above from stellar-dbt's side: the first cut split validation into a helper macro (`_test_exception_registered`) called unqualified from within `exclude_test_exceptions`'s own body. That broke with `'_test_exception_registered' is undefined` the moment stellar-dbt called `exclude_test_exceptions` qualified — because dbt resolves an unqualified macro call inside a macro's body against the **original calling node's package**, not the package where that macro is physically defined. Fixed by inlining the helper; no nested unqualified macro calls remain in either shared macro.

Also fixed a real BigQuery syntax bug in the validator caught the same way: a validation-error string used doubled-single-quote escaping (`'caller''s'`), which is the ANSI-SQL escape but not what BigQuery Standard SQL accepts. Reworded to drop the apostrophe.

Verification for both commits, against real prod data:
- all 9 of this package's own tests plus the validator compile and dry-run clean under `--no-partial-parse` (same 2 pre-existing local dev→prod relation-mapping mismatches as the original PR, unrelated to this change)
- the `sac_asset_code_matches_metadata_symbol` predicate is **byte-identical** to its pre-refactor compiled SQL; still 0 rows with the seed loaded, 1 row (the XLM SAC) with an empty exceptions table
- the validator still flags all 7 malformed-row probes with the same labels as before

CI: pre-commit, dbt-project-evaluator, diff-quality, CodeQL all green.

amishas157:
Pushed a design change (`a9da276a`, after a temp debug commit `0f896985` I've now cleaned up — sorry for the noise): `test_exception_targets()` is gone. A test now declares what it can be scoped by right in its own `config()` block:

```sql
{{ config(
    ...
    , meta={"exception_scope": {"entity_columns": [...], "allows_day_scope": bool}}
    )
}}
```

instead of in a separate dict that had to be kept in sync by hand whenever a test was wired or rewired. `exclude_test_exceptions()` no longer takes `target_key` or a registry argument — `target_key` is always `model.name`, and its declared scope is always that same test's own `config.meta.exception_scope`. `validate_test_exceptions()` builds its registry by scanning `graph["nodes"]` for test nodes in the calling project (`model.package_name`) that declare `meta.exception_scope`, instead of being handed a static dict.

This is safe across the package boundary (a downstream project calling this qualified) for the same underlying reason the PREVIOUS design needed explicit arguments: `config`/`model`/`graph` context is bound to whichever node is compiling *right now*, not to whichever package physically defines the macro executing — confirmed empirically again, not assumed.

Two dbt/Jinja behaviors surfaced while verifying this, both documented in the macro file now:
- **`graph` is only meaningfully populated once `execute` is true.** Already handled in the validator; unchanged.
- **dbt renders each node's Jinja at least twice** — an early pass where a node's own `config()` call is not yet reflected in `config.get(...)`, and a later `execute=true` pass where it is. `exclude_test_exceptions`'s meta read and validation now only run when `execute` is true; the SQL predicate itself never depended on meta (target_key/entity_columns/day_column all come from context or arguments), so it renders identically on every pass and sqlfluff's non-executing render is unaffected.

Also merged a duplicate `meta=` keyword argument in 3 files (`eho_by_ops`, `sac_asset_code_matches_metadata_symbol`, `int_token_transfer_enrichment_amount_matches_current_metadata`) that already had an inert `meta={"alert_suppression_interval": 24}` — Jinja rejects a repeated keyword outright, so this would have been a real compile break if shipped as two separate `meta=` args.

Verified against real prod data: all 9 wired tests + the validator compile under `--no-partial-parse` with no errors; `sac_asset_code_matches_metadata_symbol`'s emitted predicate is byte-identical to its pre-refactor compiled SQL, still 0 rows against prod with the seed, 1 row with an empty exceptions table; the validator's dynamically-discovered registry lists all 9 tests correctly; all 7 malformed-row probes still get flagged with the same labels. CI green.

macros/test_exceptions.sql
```diff
@@ -0,0 +1,183 @@
...
+                "exclude_test_exceptions: '" ~ target_key ~ "' has allows_day_scope="
+                ~ scope['allows_day_scope'] ~ " so day_column must "
+                ~ ('be passed' if scope['allows_day_scope'] else 'be omitted')
+            ) }}
+        {%- endif -%}
+    {%- endif -%}
+and not exists (
+        select 1
+        from {{ exceptions_relation }} as ex
+        where ex.target_key = '{{ target_key }}'
```
taha-aziz: This compares `ex.target_key` exactly, but the validator (line 131) trims it. A seed row with a trailing space would pass validation yet never match so the exception would never apply and the test keeps paging on-call. We could do `nullif(trim(ex.target_key), '')` here to match the `validator` and the `entity_column`/`entity_key` handling

taha-aziz (approved):
one small fix i'd probably do, but lgtm

macros/test_exceptions.sql
amishas157:
```suggestion
        where nullif(trim(ex.target_key), '') = '{{ target_key }}'
```

amishas157: Made change in https://github.com/stellar/stellar-dbt-public/pull/313/changes/873415293e1437e234d571a9d19c1acc4bb09f1b

